PR #1, opened 2016-12-03
Fix linux build

lefta:
Hello,

As the title states, this is a simple pull request to fix the build on Linux.

What have not been tested:
* building on Windows and Mac OS : I do not have these systems, but the changes I made should not break it.
* AVX2 and AltiVec colorspacehandler versions: My hardware uses the SSE2 version.


vadosnaprimer:
Woah! Does it actually run properly on Linux too?

lefta:
I only did a quick check, but at least it loads and runs games. As this PR only add missing includes in fsnitro.cpp and missing files in Makefile.am, it should not create any emulation problem.

vadosnaprimer:
Woops, I didn't notice it's desmume repo.

BPaden:
Build still fails if you build the gtk-glade frontend. Looks like the Makefile isn't generated.

lefta:
@BPaden not anymore ;)

PR #70, opened 2017-06-07
Let the game not because the ROM title characters of the problem can not be opened.

kongfl888:
Let the game not because the ROM title characters of the problem can not be opened.
But warning on the console!
Give some of the modified ROM a chance to run, such as MM2R NEX.
(after: a738364416119194ef2ef8b6180fdd8c50da8079)

### DIFF
--- a/desmume/src/NDSSystem.cpp
+++ b/desmume/src/NDSSystem.cpp
@@ -336,9 +336,12 @@ bool GameInfo::ValidateHeader()
 			char c = (char)header.gameTile[i];
 			if (c < 0 || (c > 0 && c < 32) || c == 127)
 			{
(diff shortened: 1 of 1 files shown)


zeromus:
Rogerman decided invalid rom titles indicated invalid roms.
You're deciding invalid rom titles are NOT invalid roms.
Which other conditions indicate invalid roms from rogerman's commit? If the title is the only validation, then we should not remove that validation. 
1. Please study whether other validation is committed, or possible.
2. Why does this rom have invalid characters? Can the NDS firmware even boot this rom with invalid characters? If the firmware can boot it and display the banner, I think we should be able to do the same thing.

kongfl888:
Clean up and close.

PR #123, opened 2018-01-08
Windows Port: Adding Pen and Touch support for touch screen devices

keelimeguy:
Noticed that a recent Windows update made my touch screen work unexpectedly with desmume, perhaps a bad driver issue or a change in how Windows interprets touch screen input as mouse events.

Regardless, I added support for some of the [WM_POINTER* family of messages](https://msdn.microsoft.com/en-us/library/hh454903(v=vs.85).aspx) which fixed the issue.

### DIFF
--- a/desmume/src/frontend/windows/main.cpp
+++ b/desmume/src/frontend/windows/main.cpp
@@ -5171,16 +5171,41 @@ DOKEYDOWN:
 		}
 		return 0;
 
+	#ifndef WM_POINTERDOWN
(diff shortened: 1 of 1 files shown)


zeromus:
Most likely windows stopped honoring our request to opt-out from the pointer handling system, and no longer implements the legacy behaviour of translating to sensible mouse input. This is no doubt due to microsoft's entire business plan revolving around selling tablets; forcing people to implement more proper touch input is a devious plan to facilitate that.

PR #147, opened 2018-06-27
Reduce RAM usage

SuuperW:
DeSmuME currently allocates 5 times the RAM required for video, in case the user selects the 5x magnification filter. This is mainly a problem when using a high 3D scaling factor.
These commits remove this excess usage of RAM by only allocating as much as is required for the currently selected filter.


zeromus:
looks OK but in the future please fix your editor to not change the formatting of unrelated code

PR #150, opened 2018-06-29
Add WiFi fix from desmume-reloaded, tweak the wait for display thread added in commit ee8fdeb

retr0s4ge:
Changes:
1- Rebased on TASVideos/desmume  (bd63ff1 tree)
2- Merge WiFi fix for desmume 0.9.12 from desmume-reloaded (commit 7548294)
3-  Forcibly wait for the display thread to draw each frame only if the user set frame skip to 0 (commit 411bcaa)

### DIFF
--- a/desmume/src/NDSSystem.h
+++ b/desmume/src/NDSSystem.h
@@ -24,8 +24,8 @@
 
 #include "types.h"
 #include "ROMReader.h"
+#include "wifi.h"
 
-class BaseDriver;
 class CFIRMWARE;
 class EMUFILE;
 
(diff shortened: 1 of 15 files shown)


w4st3d:
mudlordwhy.jpg

Seriously, why WinPCAP? I must be missing something if it must use that dependancy for sockets in particular.

zeromus:
You're missing that low level networking interfaces are required for this purpose

w4st3d:
Ah, then now it all makes sense.

mrwalters1988:
Since WinPCap is no longer being developed, can we switch the requirementes over to utilize NPCap?

zeromus:
what does whether winpcap is developed have to do with whether we use winpcap

MojoJojoDojo:
it still works with Npcap if you use winpcap API compatibility mode when you install it.

PR #152, opened 2018-07-04
Properly handle SRAM while playing movies.

SuuperW:
isMovieMode was removed because while playing a movie, the EMUFILE will always be an EMUFILE_MEMORY created for the movie, so there's no need to protect against overwriting the user's .dsv file on disk. (Writing to the EMUFILE must be allowed during movies to avoid getting an in-game error for not being able to write to the save file.)

### DIFF
--- a/desmume/src/mc.cpp
+++ b/desmume/src/mc.cpp
@@ -216,7 +216,6 @@ BackupDevice::BackupDevice()
 	fpMC = NULL;
 	fsize = 0;
 	addr_size = 0;
-	isMovieMode = false;
 
 	//default for most games; will be altered where appropriate
 	//usually 0xFF, but occasionally others. If these exceptions could be related to a particular backup memory type, that would be helpful.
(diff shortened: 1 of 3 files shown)


zeromus:
neither of these commits are good. "fix bug where movies starting from sram would crash" is leaky and the other commit is leaky and also nonsense; we cant discard the writes. I'll try to fix this the right way but then you will need to check it and maybe clean it up

SuuperW:
For the leakyness, I can look for places where BackupDevice's fpMC should be descructed.

For the second commit, one of its purposes is to NOT discard the writes; discarding writes is what the pre-commit behavior is. So I don't understand your comment on it.
(Previous behavior was to not ever write anything to SRAM while playing a movie; this would cause movies to fail if the game ever attempted to do so. New behavior is to allow said writes, though they will of course be to an in-memory EMUFILE rather than one on disk, since we don't want playing a movie to overwrite the user's default .dsv file. Basically, I am returning the behavior to how it was for version 0.9.9, the last version suitable for TASing.)

SuuperW:
@zeromus The only place where I see that BackupDevice's fpMC is not being destructed is in BackupDevice::load_movie, which I didn't change.

I'm sorry if I'm missing something that should be obvious, but how is my commit leaky?

zeromus:
ok I read it wrong. Ill just merge this and then fix it further.
`*ms` is bad karma. I just want to change how this is setup, I dont like it taking the reference to the stream.

zeromus:
I made it look more like what I want.
Please test it and investigate further.
This code is crapped up all to hell. There may be subtle lurking bugs.
I need to rewrite the garbage entirely. If you find other subtle bugs, I may do that.

PR #158, opened 2018-07-08
(Windows Frontend) Allow screen resizing in horizontal display layout.

retr0s4ge:
The right hand screen is allowed to be resized in horizontal screen layout to enable the window or full screen display to better utilize the screen area.

Changes:
1- Modify scaling, resizing and update functions to allow for new screen resizing ratio
2- Modify touch input scaling (incl. HUD editing) to adapt to different screen sizes
3- Add GUI menu for user to select the screen resizing ratio
4- Implement saving/loading settings from file similar to other settings

### DIFF
--- a/desmume/src/frontend/windows/main.cpp
+++ b/desmume/src/frontend/windows/main.cpp
@@ -717,8 +717,7 @@ void ScaleScreen(float factor, bool user)
 		else
 			if (video.layout == 1)
 			{
-				factor /= screenSizeRatio;
-				MainWindow->setClientSize((int)(video.rotatedwidthgap() * factor * 2), (int)(video.rotatedheightgap() * factor * screenSizeRatio / 2));
+				MainWindow->setClientSize((int)(video.rotatedwidthgap() * factor * 2 / screenSizeRatio), (int)(video.rotatedheightgap() * factor / 2));
 			}
 			else
 				if (video.layout == 2)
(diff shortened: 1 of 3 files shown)


retr0s4ge:
Please review and comment.
If this feature addition is OK, I can work on it further. Some improvements can still be done such as aligning the resized screen to center, handle aspect ratio differently for the resized screen.

zeromus:
This seems to have been implemented really conservatively. That's good. This code has been kludged 1000 times and it needs to be rewritten entirely from scratch before it can do anything truly flexible. There are so many things that have been broken so many times by trying to add other features, but this looks like it's probably safe. As long as you test it really really good, you can merge it

retr0s4ge:
I added a commit to the pull request to adjust the aspect ratio for the resized screen according to user setting "Maintain aspect ratio". later (tomorrow), I will add one more commit to the PR to either always vertically center the resized screen, or make that user selectable.

retr0s4ge:
I added 2 more commits to the pull request.

d17a01d for centering the resized screen vertically and
43a9734 to clear garbage from window after resizing the screen (needed only when using DirectDraw display method)

This pull request is now complete feature-wise. Please review one more time, and give OK to merge.

Thanks.

zeromus:
Looks good to me, but you may want to plug your ears before you get an avalanche of people demanding even more flexibility that it's impossible to implement without the code dissolving into a mess much like the store brand in water in an ass tissue commercial (or a total rewrite)

RareAstronaut:
Hi, this is a wonderful addition! But can we have the option to put the small screen on the bottom? Why only the top or center?

retr0s4ge:
Hi @RareAstronaut 

The top alignment was the default without me touching alignment, I added the option to center the screen but didn't add another for bottom alignment as I never thought anyone would want it.

Now that the pull request has been merged, rather than doing modifications on each user request which can result in the code being a mess (see warning above), I will note requested modifications to this feature for a while and plan accordingly.

Khold6458:
Hey, I love this feature. However, I have noticed that the DirectDraw display method is still not filling in the unused parts of the window with black. Sorry I'm not familiar with how to properly help debug this, but hopefully you can either reproduce the issue or give me instructions to figure out what the program is doing on my system.

zeromus:
Works just fine. Fiddle the code at "color is just for debug" in main.cpp to see for yourself

Khold6458:
Ah, should've caught this on my first pass but here's an update. Not working on my laptop screen, but works just fine on my external monitor. It would be great if this could be fixed, but I'm starting to suspect that this is an issue for other (Windows OS?) developers. Similar problems related to having multiple displays have popped up for me in the past.

PR #159, opened 2018-07-18
Display

SuuperW:
Display-related fixes and performance boosts.

### DIFF
--- a/desmume/src/frontend/windows/main.cpp
+++ b/desmume/src/frontend/windows/main.cpp
@@ -1996,6 +1996,12 @@ static void DoDisplay()
 	else
 		ColorspaceConvertBuffer888XTo8888Opaque<true, false>((u32*)video.srcBuffer, video.buffer, video.srcBufferSize / 4);
 
+	//some games use the backlight for fading effects
+	const size_t pixCount = video.prefilterWidth * video.prefilterHeight / 2;
+	const NDSDisplayInfo &displayInfo = GPU->GetDisplayInfo();
+	ColorspaceApplyIntensityToBuffer32<false, false>(video.buffer, pixCount, displayInfo.backlightIntensity[NDSDisplayID_Main]);
+	ColorspaceApplyIntensityToBuffer32<false, false>(video.buffer + pixCount, pixCount, displayInfo.backlightIntensity[NDSDisplayID_Touch]);
+
 	if(AnyLuaActive())
 	{
 		if(sthread_isself(display_thread))


SuuperW:
@rogerman So some games do change the backlight intensity while playing? I had not thought of this, I only thought of the user's setting on the DS' home screen.

When using DirectDraw, the backlight effect was applied prior to drawing the HUD. It seems to me that if the backlight effect is to be used, it should be applied there both for purposes of HUD and for AVI video recording.
I still think that it makes no sense to apply any backlight effect based on the home screen's display brightness setting. However given that this setting isn't normally accessible to DeSmuME users, I am fine with simply restoring the backlight effect that DirectDraw used, and having it also be used when OpenGL is selected.

This still leaves the bug where backlight intensity is not properly set after loading a savestate, mentioned in commit 6fc7161. I would have to dig into code which I am completely unfamiliar with in order to fix this, so I'll leave it for later or for someone else if they want to fix it.

zeromus:
Games can use the backlight. Read rogerman's links.
If the avi recording didnt show the result of the backlight, then that was a bug, and it needs to be fixed.  

"I still think that it makes no sense to apply any backlight effect based on the home screen's display brightness setting". Huh? There's nopart of it that doesn't make sense. The home screen sets the brightness register. We should emulate that register.

SuuperW:
@zeromus Yes, that register should be emulated if it can have any in-game effects. However, the purpose of the user's brightness setting on a physical DS is the same as the brightness setting on your monitor. There's no reason for the user to reduce the brightness there unless it has some in-game effect; in which case it would be very annoying to have to play the game with the game display permanently darkened.

The user's brightness setting on the home screen does not and cannot serve its intended purpose in the context of this emulator, even if the purpose for which some games actually make use of it can be.

zeromus:
No. Use the current register value. Whether it's the one set by the firmware, the default value we use for a no-firmware boot, or whether it's set by the game. This isn't eligible for discussion.

SuuperW:
OK, it's your project. I will not push a fix for what I consider to be a bug.
The current implementation in this PR is as you say it should be.

rogerman:
@SuuperW, what bug are you talking about?

Are you talking about the bug in the OpenGL display method where the HUD is incorrectly affected by backlightIntensity? (Note that I already acknowledged the existence of this bug beforehand.)

Look, you've already fixed the bug in commit 9ccf9ac by processing backlightIntensity only for the video frame on the CPU (which was one of the solutions that I noted beforehand).

Now -- if you're saying that the "bug" is the fact that we emulate backlightIntensity by applying a brightness effect to the video frame, then that isn't actually a bug. The feature is working as intended.

zeromus:
Choice of what to emulate is arbitrary. Emulate 666 output of GPU to LCD? Emulate brightness of screen? We have more situational awareness than you. We have a better idea what to emulate. You make a good point, though. That is one way to emulate it. In the **will never, ever, happen** case that someone requires the emulator to ignore the value of that register, it is a trivial hack to make the emulator ignore the value of that register.

fancymelonub:
Waiting for this to merge with master, can't wait to see what performance boosts will this have.

Khold6458:
I have noticed that Lua scripts that try to draw stuff are broken. Nothing appears on the screen.

zeromus:
Great, here's the never-ending cascade of stuff broken by trying to improve the windows frontend. Do you know why it didnt improve for years? Because it was impossible to improve it without rewriting it. i warned you guys. You're going to need to completely rewrite it.

Jules-A:
@Khold6458 Does anything change if you change the Display Method to/from OGL/DirectDraw?

PR #166, opened 2018-07-28
Implement basic support for multiple WiFi emulation levels

retr0s4ge:
Changes:
1- Add wifi.emulated flag to common settings
2- Don't emulate wifi unless wifi.emulated is set
3- Add a check box in windows frontend to toggle it, and read/write setting from/to ini file.


zeromus:
Looks good, but one other thing for you to pause and think about. I wonder if we might want different levels of wifi emulation -- make the setting a string (or a cryptic integer) instead of a bool? (Could be a checkbox in the GUI for now). That may be oversimplified--of course, in the long run we may need a plethora of options. But traditionally at least we've had bare-minimum-for-booting wifi emulation and the wifi "comm" emulation.

retr0s4ge:
I will add one more commit to the PR to change the setting variable name to wifi.emulationLevel instead of wifi.emulated and change its type to integer. I will leave the GUI as is for now.
Thanks for reviewing.

zeromus:
Let's let @rogerman ok it too, he may have an opinion in how it's exposed to user

rogerman:
Thanks for waiting for me to jump on board with this.

In the far flung future, it would be ideal to have a simple on/off option for WiFi, assuming that our WiFi implementation was absolutely perfect -- DeSmuME would be a perfect network citizen that doesn't bog down either the host's network adapter or the network itself, as well as DeSmuME using minimal CPU cycles for its WiFi emulation.

So I agree that there should be multiple levels of WiFi support, but only because WiFi is still very much under development and that we don't know everything about it. To be safe, the user should be able to choose between different WiFi emulation levels so that they would have better control over using multiple imperfect solutions. But rather than using a type int, it would be better to use typed enums instead.

So now the question is: What WiFi levels should we support?

Here is what I think we should use to start with, and also the philosophy behind each WiFi level choice.

enum WifiEmulationLevel
{
WifiEmulationLevel_Off = 0, // Emulates all WiFi features that are not dependent on either the host's network adapter or pcap. This option basically emulates the fact that the NDS cannot find a network to connect to. Can be useful for debugging. Can also be useful for performance purposes if the user does not want to use any additional CPU cycles for WiFi emulation, or consume network bandwidth with emulated WiFi packets.

WifiEmulationLevel_Normal = 10000, // Emulates all WiFi features that can be reasonably implemented without overly stressing the host hardware or the network. This emulation level should work for most uses cases, but is not required to work for all use cases. The programmer is allowed to use their discretion to create performance/accuracy tradeoffs in the WiFi emulation as necessary.

WifiEmulationLevel_Compatibility = 65535 // Emulates all WiFi features possible without any compromises whatsoever. This emulation level should attempt to work for all use cases. The programmer is NOT allowed to create performance/accuracy tradeoffs and must always go for accuracy in every detail, even if said accuracy would reduce the overall performance of the host hardware or network.
};

If necessary, it would be possible to add additional WiFi emulation levels that are variants of WifiEmulationLevel_Off and WifiEmulationLevel_Normal. But I think that the three levels I listed are a good place to start.

retr0s4ge:
Great idea, I will add another commit tomorrow for the changes.
Thanks for reviewing.

retr0s4ge:
@rogerman Done.

retr0s4ge:
@zeromus Please give OK to merge. Thanks.

zeromus:
ok merge it. in general, dont use c++11 features

rogerman:
Just to add -- assume that the lowest-common-denominator compiler is compatible with C99 and C++03. However, platform-specific extensions using #ifdef are okay, just try not to go overboard with them.

Issue #164, opened 2018-07-24
Waiting for the display thread to draw each frame while vsync is ON slows emulation down if monitor refresh rate is lower than 60


retr0s4ge:
Waiting for the display thread to draw every frame (currently activated only if frameskip is 0) while vsync is ON slows the emulation down if the monitor's refresh rate is lower than 60.
ex. Internal fps is capped at 48 if the monitor has 48Hz refresh rate

Each of these options work correctly if the other is not activated.

vsync option alone allows internal fps to change freely depending on emulation and 3D settings (not capped at monitor refresh rate), yet the final display rate is synchronized with the monitor and no screen tearing is visible.

Setting frame skip to zero alone activates the wait for display thread to draw each frame, which works as intended, and the fps is not capped at monitor refresh rate.

@SuuperW

SuuperW:
It sounds like it is working as intended.
If the user requests both vsync and zero frame skipping, he will expect the output video to be synchronized with the monitor and every video frame to be visible. This is what happens, according to your description.

You cannot logically run at a higher fps than your display's fps with zero frame skipping on that display.

zeromus:
It's logical. you could argue that frameskipping is designed to help reach target speed (60fps) NOT reach the vsync'd rate. In that case it should have no effect when the monitor is running at 48hz. I'm sympathetic to a user who wants to get what amounts to a good triple buffered solution for 60fps game running without tearing at 48hz. That said, I think you're all nuts for spending any time on this junky old frontend and every time you do change one thing it will become increasingly implausible that youve done it without breaking something else or ever fix anything else again.

Jules-A:
While this may be working as intended, turbo while under 60 is severely affected. Turbo at 40 fps makes fps go to around 50 but looks like it's skipping most frames. Before the initial changes by SuuperW, turbo at 45fps (even with newer builds being 64bit, they are 15% slower than the old Nighly's but better than 30% diff between 32bit builds) would bring fps to 105. Basically with the changes, turbo is unusable if your FPS is just above 60fps or lower. If I lower the GPU factor to 3 (what I normally play at) which gets me to 66fps (uncapped) on the old Nightly turbos to 125fps while the newer builds are normal 61fps and turbos to 83 fps but again looks like it's skipping too many frames.

I know this is really a separate issue, I was planning to write a detailed issue but haven't had the time. I thought this might be helpful as it's still related.

zeromus:
[...]

PR #168, opened 2018-07-30
(Windows Frontend) Add an option to display menu in fullscreen mode.

retr0s4ge:
This should eliminate the need to keep toggling fullscreen mode on/off to change settings.

### DIFF
--- a/desmume/src/NDSSystem.cpp
+++ b/desmume/src/NDSSystem.cpp
@@ -102,7 +102,7 @@ int TotalLagFrames;
 
 TSCalInfo TSCal;
 
-WifiEmulationLevel wifiEmulationLevel;
+WifiEmulationLevel wifiEmulationLevel = WifiEmulationLevel_Off;
 
 namespace DLDI
 {
--- a/desmume/src/NDSSystem.h
+++ b/desmume/src/NDSSystem.h
@@ -709,8 +709,7 @@ extern struct TCommonSettings
 
 	std::string run_advanscene_import;
 
-};
-extern TCommonSettings CommonSettings;
+} CommonSettings;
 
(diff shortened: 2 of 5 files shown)


retr0s4ge:
As a useful side effect while this setting is enabled, WDWM's triple buffering and vsync will be applied in fullscreen mode on windows 8 and 10 (And supposedly, windows 7 if Aero theme is enabled).
So, this can also be used as a workaround for #164

zeromus:
solve it always by making the pseudo-fullscreen window 1 pixel smaller, or something like that. see "enable windows fullscreen hacks" in bizhawk, although we're far from cutting edge in bizhawk OR desmume. in both cases we have mountains of legacy cruft to deal with so we can't really incorporate modern best practices.......whatever those are.........

retr0s4ge:
Thanks for the pointers, I will check that as a primary solution for #164 

If we disregard the side effect, what's your opinion on having the option to display the menu in fullscreen mode ?

I think the menu has a fairly low profile, and can be convenient in case the user wants to change a setting without leaving fullscreen.

zeromus:
windows has no such feature. you have to code it yourself.

I don't care if there's a menu in fullscreen mode. Some people might. Dont do this just as a lark, or as a workaround when there's the other workaround I mentioned. If someone wants this, then it's worth doing. That is, if working on the desmume frontend at all rather than rewriting it is worth doing in the first place. I don't think so, but do what you want with your time. Just beware the first time someone actually rolls up their sleeves and rebuilds the windows frontend, all your code will probably be deleted.

One thing bizhawk is kind of cutting edge in (or at least comfortably near the front of the pack) is in options for getting work done in windows, which includes this. Consequently, these are all options people have wanted:

image

PR #175, opened 2018-08-11
(Windows Frontend) Add Fullscreen Options And Window Sizing Fixes

### LINKED ISSUE #164: Waiting for the display thread to draw each frame while vsync is ON slows emulation down if monitor refresh rate is lower than 60

retr0s4ge:
Added the following Options:
1- Windowed Fullscreen mode (fixes #164), assuming user turns off vsync option in desmume as DWM will handle that
2- Show menu in full screen mode, and fixed settings changes (from menu or hotkey) that break window size
3- Auto-hide cursor option (Full screen mode)

Fixes:
1- Screen gap is no longer visible in horizontal layout when the window is resized when using opengl display method if the layout was changed from vertical to horizontal without restarting the emulator
2- No longer possible to rotate the video using hotkey in horizontal layout
3- Changing layout from vertical to horizontal while video is rotated and windowsize = 0 no longer increases window size
4- Changing screen layout while screensizeratio is > 1 and windowsize = 0 no longer increases window size
5- Repeatedly entering and exiting full screen mode while windowsize = 0 no longer increases window size
6- Pressing the menu key has no effect if the main window has no menu

Unrelated change: add *.VC.opendb to gitignore (generated by VS when solution is open)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 *.user
 *.suo
 *.VC.db
+*.VC.opendb
 *.o
 .deps
 *.dirstamp
--- a/desmume/src/frontend/windows/hotkey.cpp
+++ b/desmume/src/frontend/windows/hotkey.cpp
@@ -347,12 +347,10 @@ void HK_AutoHoldClearKeyDown(int, bool justPressed) {
 }
 
 extern VideoInfo video;
-extern void doLCDsLayout();
+extern void doLCDsLayout(int videoLayout);
 void HK_LCDsMode(int)
 {
-	video.layout++;
-	if (video.layout > 2) video.layout = 0;
-	doLCDsLayout();
+	doLCDsLayout(video.layout + 1);
 }
 
 extern void LCDsSwap(int);
--- a/desmume/src/frontend/windows/main.cpp
+++ b/desmume/src/frontend/windows/main.cpp
@@ -514,6 +514,8 @@ bool frameAdvance = false;
 bool continuousframeAdvancing = false;
 
 unsigned short windowSize = 0;
+bool fsWindow = false;
+bool autoHideCursor = false;
 
 float screenSizeRatio = 1.0f;
 bool vCenterResizedScr = true;
@@ -692,8 +694,12 @@ void ScaleScreen(float factor, bool user)
 {
 	if(user) // have to exit maximized mode if the user told us to set a specific window size
 	{
-		bool maximized = IsZoomed(MainWindow->getHWnd())==TRUE;
-		if(maximized) ShowWindow(MainWindow->getHWnd(),SW_NORMAL);
+		HWND hwnd = MainWindow->getHWnd();
+		if (IsZoomed(hwnd) || fsWindow)
+		{
+			RestoreWindow(hwnd);
+			windowSize = factor;
+		}
 	}
 
 	if(windowSize == 0)
@@ -977,6 +983,8 @@ const u32 DWS_DDRAW_HW = 32;
 const u32 DWS_OPENGL = 64;
 const u32 DWS_DISPMETHODS =  (DWS_DDRAW_SW|DWS_DDRAW_HW|DWS_OPENGL);
 const u32 DWS_FILTER = 128;
+const u32 DWS_FS_MENU = 256;
+const u32 DWS_FS_WINDOW = 512;
 
 static u32 currWindowStyle = DWS_NORMAL;
 void SetStyle(u32 dws);
@@ -985,15 +993,77 @@ static u32 GetStyle() { return currWindowStyle; }
 static void ToggleFullscreen()
 {
 	u32 style = GetStyle();
-	style ^= DWS_FULLSCREEN;
-	SetStyle(style);
+	HWND hwnd = MainWindow->getHWnd();
+
 	if(style&DWS_FULLSCREEN)
-		ShowWindow(MainWindow->getHWnd(),SW_MAXIMIZE);
-	else ShowWindow(MainWindow->getHWnd(),SW_NORMAL);
+		RestoreWindow(hwnd);
+	else ShowFullScreen(hwnd);
 }
 //---------
 
+void SaveWindowSizePos(HWND hwnd)
+{
+	SaveWindowPos(hwnd);
+	SaveWindowSize(hwnd);
+	WritePrivateProfileInt("Video", "Window Size", windowSize, IniName);
+}
 
+void RestoreWindow(HWND hwnd)
+{
+	fsWindow = false;
+	u32 style = GetStyle();
+
+	if (style & DWS_FULLSCREEN)
+	{
+		style &= ~DWS_FULLSCREEN;
+		SetStyle(style);
+		ShowWindow(hwnd, SW_NORMAL);
+
+		windowSize = GetPrivateProfileInt("Video", "Window Size", 0, IniName);
+		WndX = GetPrivateProfileInt("Video", "WindowPosX", CW_USEDEFAULT, IniName);
+		WndY = GetPrivateProfileInt("Video", "WindowPosY", CW_USEDEFAULT, IniName);
+
+		RECT rc;
+		rc.left = WndX;
+		rc.top = WndY;
+		rc.right = rc.left + GetPrivateProfileInt("Video", "Window width", GPU_FRAMEBUFFER_NATIVE_WIDTH, IniName);
+		rc.bottom = rc.top + GetPrivateProfileInt("Video", "Window height", GPU_FRAMEBUFFER_NATIVE_HEIGHT * 2, IniName);
+
+		AdjustWindowRect(&rc, GetWindowLong(hwnd, GWL_STYLE), true);
+		SetWindowPos(hwnd, 0, WndX, WndY, rc.right - rc.left, rc.bottom - rc.top, SWP_NOOWNERZORDER | SWP_NOZORDER);
+	}
+	else
+		ShowWindow(hwnd, SW_NORMAL);
+}
+
+void ShowFullScreen(HWND hwnd)
+{
+	SaveWindowSizePos(hwnd);
+
+	u32 style = GetStyle();
+	style |= DWS_FULLSCREEN;
+	SetStyle(style);
+
+	if (style & DWS_FS_WINDOW)
+	{
+		windowSize = 0;
+		fsWindow = true;
+
+		HMONITOR monitor = MonitorFromWindow(hwnd, MONITOR_DEFAULTTONEAREST);
+		MONITORINFO info;
+		info.cbSize = sizeof(MONITORINFO);
+		GetMonitorInfo(monitor, &info);
+		int monitor_width = info.rcMonitor.right - info.rcMonitor.left;
+		int monitor_height = info.rcMonitor.bottom - info.rcMonitor.top;
+
+		SetWindowPos(hwnd, 0, info.rcMonitor.left, info.rcMonitor.top, monitor_width, monitor_height, SWP_NOOWNERZORDER | SWP_NOZORDER);
+	}
+	else
+	{
+		fsWindow = false;
+		ShowWindow(hwnd, SW_MAXIMIZE);
+	}
+}
 
 static HMENU GetMenuItemParent(UINT itemId, HMENU hMenu = mainMenu)
 {
@@ -1213,6 +1283,8 @@ void UpdateWndRects(HWND hwnd, RECT* newClientRect = NULL)
 	POINT ptClient;
 	RECT rc;
 
+	GapRect = {0,0,0,0};
+
 	bool maximized = IsZoomed(hwnd)!=FALSE;
 
 	int wndWidth, wndHeight;
@@ -1421,20 +1493,24 @@ void LCDsSwap(int swapVal)
 	WritePrivateProfileInt("Video", "LCDsSwap", video.swap, IniName);
 }
 
-void doLCDsLayout()
+void doLCDsLayout(int videoLayout)
 {
 	HWND hwnd = MainWindow->getHWnd();
+	u32 style = GetStyle();
+	bool maximized = IsZoomed(hwnd) || fsWindow;
+	if (maximized)
+		RestoreWindow(hwnd);
 
-	bool maximized = IsZoomed(hwnd)==TRUE;
+	if (videoLayout > 2) videoLayout = 0;
 
-	if(maximized) ShowWindow(hwnd,SW_NORMAL);
-
-	if(video.layout != 0)
+	if(videoLayout != 0)
 	{
 		// rotation is not supported in the alternate layouts
 		if(video.rotation != 0)
 			SetRotate(hwnd, 0, false);
 	}
+
+	video.layout = videoLayout;
 
 	osd->singleScreen = (video.layout == 2);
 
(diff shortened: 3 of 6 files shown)


zeromus:
Looks fine to me

zeromus:
No, users have to choose whether to use that option or not. There are pros and cons depending on the user's GPUs and monitors and preferences. I dont know if "exclusive mode" is the correct terminology though. I thought folks had settled on "fullscreen" and "windowed fullscreen"

rogerman:
Well it doesn't really matter either way because Windows DeSmuME has never used exclusive fullscreen mode in the first place. It's always used some form of windowed mode. So until we start actually using some form of exclusive fullscreen mode, having an option for "Non-exclusive Mode" is pointless.

zeromus:
Some video drivers put the program into exclusive fullscreen mode when the program naively attempts to enter windowed fullscreen mode and foolishly makes the window the exact size of the screen. For those drivers, you must make the window one pixel than the size of the screen.

Do I really need to keep going? The less you have to know about this, the happier you'll be, trust me.

Anyway  i assume something like that is what retr0s4ge did; otherwise what's the difference right now even between those two modes? I agree it would be good to be able to genuinely do exclusive fullscreen too, but in actuality this issue is a total clusterfuck and in the perfect world there are even 20 times more options than we have. Go look at retroarch for an evolved example.

rogerman:
Well this code certainly won't help with that. See ShowFullScreen(). It resizes the window to be exactly the size of the monitor. Therefore, the "Non-exclusive Mode" option will fail to give the user consistent control over the actual fullscreen mode, making it useless.

And for the record, I probably know more than you think I do about this stuff. OS X has been doing this kind of window compositing stuff before Windows Vista and DWM ever became a thing. Creating an ideal drawing setup based on OS version and states is not complicated. Dealing with video drivers and different OS versions trying to do their own thing with borderless fullscreen windows is old hat. And just to humor you, I did look at RetroArch, and what it does is extremely simplistic. All RetroArch does is attempt to set the DWM compositor enable state and nothing more. That is the one and only option that RetroArch has related to this issue. I think you are overblowing the complexity of it all. In a perfect world, there are no options -- the app should know exactly what the best thing is to do based on state.

Now... if exclusive fullscreen ever comes to DeSmuME, then the "Non-exclusive Mode" option would make sense. But since such a feature does not exist, it makes no sense to have the option, since the option would make the user think that they might have control, when in fact they don't.

zeromus:
Exclusive mode makes it fill the entire monitor. Non-exclusive mode maximizes it, but the frame has 1px remaining (see SetStyle(), I think). Therefore, there is a difference. 

The app doesn't know "the best thing to do" because at the minimum, there is some tradeoff between latency and performance. There are also considerations concerning monitor refresh rates, synchronization on stereo 3d displays, the difficulties in determining which screen's parameters should be used for a given windows position. These all affect which vsyncing methods are functional or unreliable or unavailable. There isn't any way to do this right, and best isn't good enough. All we can do is offer a grotesque flabby mountain of options.

As for retroarch, these are all related in how the vsyncing, frame submission, and monitor associations work, and I just counted a half dozen options related to that. Assess whichever minute fraction of retroarch's options you wish, that's your prerogative as guy who woefully undestimates the colossal mess of this subject.

retr0s4ge:
Thanks for reviewing and commenting.

I fixed the bug where the cursor was being hidden when over the menu, I also increased the time before auto-hiding to 10 seconds.

Regarding the fullscreen implementation:
XP was the last windows version to have applications work unrestricted in truly exclusive fullscreen mode (maybe count windows 7 in if the user disables the aero theme).

Now DWM handles all windows, forces vsync regardless of application settings, causes a reduction in frame rate and increases input lag.
So in return, windows GPU drivers use their own "hacks" to "improve" performance, and most will switch any application whose client area covers the whole screen to exclusive fullscreen mode, thus bypassing DWM. (This is almost guaranteed nowadays)
The intention is to improve frame-rate and reduce input latency at the cost of introducing some screen tearing which can be fixed if the user enables vsync inside the application.
However, there are some advantages when having DWM handle the fullscreen window such as free (and apparently forced) triple buffering and complete elimination of fullscreen mode flickering that occurs on some machines.

In order to force the application to NOT go to exclusive fullscreen mode regardless of the GPU model, the client area should be at least 1px short from covering the whole screen.
In other words, some part of the non-client area of the window has to stay on the screen so that DWM handles it, whether it is a border or menu or caption, etc... (DWM will also handle all windows if any part of the taskbar or desktop is visible).

The choice when deciding how to go fullscreen actually comes down to:
1- Escape DWM (aka: exclusive fullscreen mode)
2- Have DWM handle the window (aka: fullscreen window or non-exclusive fullscreen mode)

The user should be able to choose whether DWM handles the window or not (some might not mind input lag but hate tearing, others will want the opposite).

Going exclusive fullscreen is easy, just set WS_POPUP style and maximize the window, the GPU driver does the rest.

The problem happens if you want non-exclusive fullscreen, the simple way is to set WS_POPUPWINDOW style and maximize the window.
Sadly, some GPU drivers will not draw the border and still switch to exclusive fullscreen mode anyway.

To guarantee that this doesn't happen, rather than maximizing the window, one would resize the window to the monitor size with the client area being at least 1px smaller.

There are multiple ways for doing this, but I find the easiest way is to do the following:
(this is what's done in the commit)
1- Set the WS_POPUPWINDOW style, which leaves a small border
2- Get the monitor size
3- Resize the window to fit the monitor size

Note that step 3 says "Resize the window" rather than "Resize the client area", as the client area will be automatically smaller because of the thickness of the border set in place by the WS_POPUPWINDOW style.

The downside to doing this is that Windows will still see the fullscreen window as being "normal" rather than "maximized", so some extra handling is required to "restore" the window. This is done in the RestoreWindow function.

retr0s4ge:
Please check and give OK to merge or clarify further requirements.
Thanks.

rogerman:
Hi @retr0s4ge, sorry for the delayed reply. Ironically, my Windows 10 decided to automatically do an update, which then rendered it unable to boot itself! After many many many hours of troubleshooting, I finally bit the bullet and pulled a "Reset this PC". I've been busy reinstalling stuff so I was only able to test your new changes just now.

• Commit 775e3ae successfully resolves the issues with the auto-hiding cursor in combination with the menu bar. Also, the auto-hide cursor timing at 10 seconds seems to work very well. It is very good with keeping consistent visibility of the cursor when using the menu a lot or emulating the stylus sporadically. At a 10 second auto-hide, the cursor disappears only when I'm good and done with it. Works great!

• However, I have discovered another bug. If the menu is hidden in fullscreen; use into fullscreen and press the Alt key. The emulator will pause as if the menu is being accessed. This shouldn't happen, since the user can't access the menu anyways, nor can the user use keyboard shortcuts while the menu is hidden.

The UI becomes even more inconsistent when the user presses the Alt key (which pauses the emulation), and then presses the Return key (not referring to an Alt-Return keystroke combination, but Alt and Return being used separately.) Now you have the window menu appearing out of nowhere, where Minimize and Close work, but the other options don't work (greyed out or otherwise).

When the menu is hidden in fullscreen while the user is using fullscreen mode, you should prevent the Alt key from accessing any menus.

• As for the stuff about the user-selectable exclusive vs. non-exclusive fullscreen modes... just do what you want. I won't stop you guys from adding a superfluous option that will make users think they are receiving a tangible benefit, when in fact the DeSmuME Windows code does not support it. This is the last time I will say that the Windows port code prevents exclusive mode from giving you the performance and input lag improvements that you think you're going to get. To truly get those benefits, the entire Windows port video code needs to be rewritten from the ground up with modern graphics API in combination with cooperating with the DeSmuME GPU API that I've written over the past two years in order to address these very issues in the Cocoa port.

retr0s4ge:
Thanks for approving.

Just to clarify, the main purpose I wrote this for is to eliminate screen tearing in fullscreen mode without having vsync turned on in the settings.
This allow emulation to run full-speed on devices with monitor refresh rate lower than 60 without tearing or visibly skipping frames. (Basically have vsync + triple buffering applied outside desmume).

This already functions as intended as demonstrated here:
https://youtu.be/D47ZyoBbL00

When vsync + triple buffering are applied, some minor input lag is introduced, and because we rely on DWM to handle the fullscreen window to acheive that, the frame rate drops a bit compared to not doing any of that.

I wouldn't be wrong saying the following:
Turning on the option currently named "non-exclusive fullscreen mode" reduces the overall performance compared to not having it on. However, screen tearing among other issues will be eliminated.

So, I am not claiming any of desmume's fullscreen options increases performance or reduces input lag. Actually, I am stating the opposite.

As for the naming, I have initially chosen "non-exclusive fullscreen mode" to reflect that DWM is guaranteed to handle the window in fullscreen mode if that setting is selected.

Perhaps renaming that to something else resolves all issues ?

rogerman:
No need to clarify why you did what you did or make a demonstration about different v-sync behavior - trust me, I know the drill. And no, you don't need to rename stuff for better clarity because the naming shows the intention of the programmer.

My argument against having the option in the first place is threefold:
1. The programmer's intention is that enabling "Non-exclusive Mode" will force DeSmuME to cooperate with DWM. However, the converse of this is that disabling "Non-exclusive Mode" will cause DeSmuME to work in exclusive fullscreen mode. Unfortunately, neither of these options are guaranteed to actually work, making this a deceptive option for the user.
2. Due to the way Windows DeSmuME is designed at the moment, it is simply best to assume that DWM is the desired mode for all situations. Exclusive fullscreen does not provide any significant advantages in either video performance or input lag when compared to just cooperating with DWM. In order to gain those advantages, Windows DeSmuME's video and input handling code will need to be completely reworked. In other words, Windows DeSmuME itself has far more influence over video performance and input lag than DWM does. In short, you’re better off assuming the use of DWM under all circumstances because the performance and input lag differences are inconsequential.
3. On a lesser note, if DeSmuME ever does take exclusive fullscreen mode, then there is a bug with many drivers in which trying to switch out of fullscreen will cause DeSmuME to get stuck in exclusive fullscreen, but with the actual viewport being the size of the original window. (Nvidia drivers seem to be a lot better about doing the right thing, but Intel and AMD drivers tend to cause the bug.) This situation forces the user into using End Task on DeSmuME in order to regain access to the desktop. However, this is something I'm not holding against you, since this particular set of commits is not intended to address this particular bug, and this bug has already existed in the past.

So let's go over the issues in a little more detail.

Windows DeSmuME essentially uses two video display modes: DirectDraw7 and OpenGL.

Regardless of Windows version or GPU driver, DirectDraw7 will always cooperate with DWM if present. Therefore, in Windows 8 and 10, DirectDraw7 will always use DWM. In Windows Vista and 7, DirectDraw7 will always use DWM as long as Aero is enabled. But if Aero is disabled, DirectDraw7 will not use DWM. So in short, DirectDraw7 has the most consistency when it comes to V-sync because it is always at the mercy of DWM. In fact, DeSmuME's V-sync option doesn't even work on DirectDraw7 -- it only works with OpenGL. In addition, your new "Non-exclusive Mode" option is useless with DirectDraw7 because DirectDraw7 always cooperates with DWM regardless.


Now things start getting interesting when using the OpenGL video display mode. Drivers have a lot more influence on what goes on here, so let’s look at some of those things:

• If the window content area is exactly the same size as the monitor and if the window content area is completely opaque, then the driver MIGHT automatically switch to exclusive fullscreen mode as long as no other things are visible in front of the window. However, the driver MIGHT also choose to continue cooperating with DWM.
- In Windows 8 and 10, drivers tend to cooperate with DWM more often than not, but there are still a few drivers (usually AMD or Intel) that will still attempt to take exclusive fullscreen anyways.
- In Windows Vista and 7, drivers tend to take exclusive fullscreen mode more often than not, but there are still a few drivers that will continue cooperating with DWM anyways. However, the user can turn off Aero to give the driver a hint to use exclusive fullscreen mode, and so turning off Aero virtually guarantees that the driver will go exclusive fullscreen. Unfortunately, Windows 8 and 10 do not have the option to turn off Aero or disable DWM at the system level.
- In short, disabling “Non-exclusive Mode” does not always guarantee that DeSmuME will use exclusive fullscreen mode.

• In Windows Vista and 7, if the entire window area covers the monitor, and as long as all window areas (including borders) are completely opaque, then the driver MIGHT automatically switch to exclusive fullscreen mode as long as no other things are visible in front of the window. However, in most cases, drivers will usually just cooperate with DWM at this point. Therefore, in Windows 7/Vista, enabling “Non-exclusive Mode” will usually, but not always, have the intended effect.
- This scenario does not apply to Windows 8 and 10 because the driver will cooperate with DWM in ALL circumstances because even single-pixel wide window borders are transparent. Therefore, enabling “Non-exclusive Mode” on Windows 8/10 is guaranteed to work.

So clearly, leaving things up to the driver leads to inconsistent behavior. The only way to truly guarantee that DeSmuME uses exclusive vs non-exclusive fullscreen modes is to explicitly request such modes using Direct3D API. Unfortunately, DeSmuME doesn’t use Direct3D -- it only uses DirectDraw7 and OpenGL, and so we can’t do this at this time.


Finally, let’s discuss V-sync.

- As stated before, the V-sync option only works with OpenGL, but only if the driver chooses to acknowledge it. If the user goes into fullscreen mode (could be either one), some drivers (usually Nvidia drivers) can opt to ignore the user’s V-sync option because it knows that its cooperating with DWM, which has its own built-in V-sync. In my personal testing, ignoring the user’s V-sync setting happens most often under Windows 10, and never under Windows 7. However, there could be some outliers to this rule that have escaped my testing.

- Assuming that the user is running OpenGL video display mode and that driver is respecting the user’s V-sync setting, Windows DeSmuME has an issue where its FPS will be capped to the refresh rate of the monitor, as noted in issue #164. This is the fault of the Windows DeSmuME video display thread blocking itself on V-sync.

- I am assuming that your video was made while you were running OpenGL video display mode without V-sync, probably under Windows 7 too. If that’s the case, then why not use DirectDraw HW? It will use DWM to V-sync your video under all circumstances. The frame capping issue from #164 only affects OpenGL, but not DirectDraw, and so DirectDraw will give you the solution that you’re looking for.

retr0s4ge:
That video was taken on windows 10 build 1803, and using a gtx970m GPU.
I also had to force vsync off in the application profile in the nvidia control panel as the nvidia driver always enabled vsync regardless of the application setting even though the selected setting clearly says "use application settings"

So in this situation, either vsync is forced by the driver resulting in fps cap or completely disabled leading to tearing. The only viable option was relying on DWM for the vsync and triple buffering.

I am leaning to change the name of the option to either "Bordered Window" which is descriptive of what is done and doesn't imply anything else or "Fullscreen hacks" which I was reluctant to use for containing the word "hacks".

rogerman:
Well if that's the case, then you should name it "Force Windowed Fullscreen". The term "Windowed Fullscreen" is a mainstream term understood by gamers, and "Force" implies that the user really wants DWM. This option name should accurately reflect what is happening on Windows 8 and 10, although this behavior is not 100% guaranteed on Windows Vista and 7 but is still true in the vast majority of cases.

And again, this option is only relevant for OpenGL display mode. The DirectDraw display modes will remain unaffected since they always run under DWM.

retr0s4ge:
I have cleared all points that were raised in the discussion.
Please recheck and merge if there are no further issues.
Thanks.

Issue #178, opened 2018-08-20
Request: Cheat List window improvements.


ReyVGM:
Hey guys.

Would it be possible to modify the cheat list window a bit with the following improvements?

**PENDING:**

-Ability to highlight multiple cheats to delete them, instead of deleting them one by one. 
When creating my own cheats, I usually have to add 10 or 20 addresses for testing, but deleting the bad ones afterwards is tedious since I can only delete one by one, and can't even use the keyboard for it.

-Ability to move cheats to the top or bottom. 
When a cheat list is too big, and you add a new cheat or if I'm making my own cheats, the new codes will always appear in the bottom. Which means that if I'm constantly modifying the value, or enabling/disabling the cheat, I have to scroll all the way to the bottom. It would be nice if I could just move those cheats to the top to avoid scrolling every time.

**ALREADY IMPLEMENTED:**

-Ability to enable / disable all available cheats.
A welcomed feature would be the option of clicking on a box or something to check-mark all available cheats (and uncheck them too). This is specially needed when opening the Cheats Database, since you have to click each cheat one by one to add it to the emulator.

ReyVGM:
Also, the cheat descriptions get cut off by the useless thing:

![new bitmap image](https://user-images.githubusercontent.com/7338867/44323576-86aa0c00-a420-11e8-9231-ee1c91907f87.PNG)

zeromus:
So is this fixed or what

zeromus:
let's say its fixed unless someone tells me otherwise

PR #181, opened 2018-08-25
Cheats

SuuperW:
Implement request #178, along with a couple other cheat list related features.

I'm not sure if all the commits here are correct. Once my previous PR is merged I will fix this PR if necessary.


SuuperW:
The AppVeyor build failed, but the code builds just fine on my machine (Win10, VS2017).
The error is "cheatSystem.cpp(82): error C2039: '_Make_iterator_offset': is not a member of 'std::vector<CHEATS_LIST,std::allocator<_Ty>>'".
I do not see why it can't find the _Make_iterator_offset method. Does anyone else know?

zeromus:
your move() implementation is really weird. Find a more normal way of doing it. That's some kind of internal implementation detail you've used on the STL type.

Also, I know it's annoying, but when you've got multiple PR in flight like this, you should probably base the branch on origin/master, rather than suuperw/master, so the changes can be considered in isolation rather than along with all your other edits to master. I'm not asking you to change it though. However, something small like this could get hung up waiting for the bigger PR to get done

SuuperW:
I've updated the move method, and the automatic build is successful now.

SuuperW:
I've fixed the PR.

JTR425:
sorry if it was extremely obvious but how do i move cheats?

zeromus:
drag them, or edit the .dct file

JTR425:
do i have to hold down say shift or something when dragging? also how do i edit the dct file. sorry if these are bad questions i only just downloaded desmume and have no experience in this

JTR425:
sorry to be a bother i figured it out thanks for the help anyways

PR #187, opened 2018-09-03
Movies

SuuperW:
Use a movie's emulation/firmware settings, and support movies starting from a savestate.


zeromus:
I can't really evaluate this, I havent dealt with any of this in too long. I'll merge it soon.

PR #188, opened 2018-09-05
Paths

SuuperW:
Properly save and use paths.

### DIFF
--- a/desmume/src/firmware.cpp
+++ b/desmume/src/firmware.cpp
@@ -603,11 +603,9 @@ std::string CFIRMWARE::GetExternalFilePath()
 {
 	std::string fwPath = CommonSettings.Firmware;
 	std::string fwFileName = Path::GetFileNameFromPathWithoutExt(fwPath);
-	char configPath[MAX_PATH] = {0};
-	path.getpath(path.BATTERY, configPath);
-	if (configPath[strlen(configPath)-1] == DIRECTORY_DELIMITER_CHAR)
-			configPath[strlen(configPath)-1] = 0;
-	std::string finalPath = std::string(configPath) + DIRECTORY_DELIMITER_CHAR + fwFileName + FILE_EXT_DELIMITER_CHAR + FW_CONFIG_FILE_EXT;
+
+	std::string configPath = path.getpath(path.BATTERY);
+	std::string finalPath = configPath + DIRECTORY_DELIMITER_CHAR + fwFileName + FILE_EXT_DELIMITER_CHAR + FW_CONFIG_FILE_EXT;
 
 	return finalPath;
 }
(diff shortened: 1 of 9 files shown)


zeromus:
Can you describe what bugs were fixed by this?

SuuperW:
commit 407b01b
DeSmuME would not use the saved "Battery" folder location when exporting a sav.

commit aaad5e6
DeSmuME would not take the user to the saved "Screenshots" folder location when saving a screenshot, unless the file name it chose already existed.
DeSmuME would not save the selected folder locations when importing/exporting a sav, saving a screenshot, recording an AVI or WAV, setting external firmware location, or setting the microphone sample file.

zeromus:
I'm sorry, I don't understand you at all. Please try writing it like you were reporting the bug with detailed repro steps.

SuuperW:
(with default settings)

Not using saved path settings:
1) Open a ROM via the Open ROM dialog (not via recent ROMs list)
2) Select File -> Export Backup Memory...
--Expected behavior: Dialog opens with the Battery folder (assuming default settings) as the initial directory.
--Actual behavior: Dialog opens with the ROMs folder as the initial directory. [This is because DeSmuME does not attempt to use the value for PathSettings->Battery from the .ini file.]
or 2) Select File -> Save Screenshot As...
--Expected behavior: Dialog opens with the Screenshots folder (assuming default settings) as the initial directory.
--Actual behavior: Dialog opens with the ROMs folder as the initial directory. [This is because DeSmuME sets the open file dialog's "lpstrFile" field to a path+file which does not exist, so the dialog falls back to whatever directory was last opened in a file dialog.]

Not saving path settings:
1) Select File -> Import Backup Memory... (this will open a dialog with the Battery folder as the initial directory)
2) Import a .sav from any other folder
3) Close and re-open DeSmuME
4) Select File -> Import Backup Memory...
--Expected behavior: The initial directory of the dialog should be the location of the file selected in step 2.
--Actual behavior: The initial directory of the dialog is the Batter folder.
(Similar behavior for other dialogs which use saved settings to determine their initial directory.)


While doing this I discovered that some other open file dialogs (e.g. Record AVI) have the same problem as the Save Screenshot dialog. I will fix these and commit.

zeromus:
Ideally, there is a memory for each dialog, which is populated only once with the defaults. 
so the pseudocode is this

```
METHOD 1
if dialog's path is empty
  set path to configured default
open dialog with path
remember that path for next time
```

there's another way, though. I'm not sure which is preferable. 

```
METHOD 2
set path to configured default
open dialog with path
```

Based on your 2nd point, it seems you prefer METHOD 2.

Can you clarify which method you're using?
I'm evaluating this topic for bizhawk as well.
@vadosnaprimer votes for METHOD 2

PR #190, opened 2018-09-07
Savestates

SuuperW:
First two commits allow some old savestates to be loaded with no readily-apparent problems.
3rd and 4th change the structure of savestates, to address issues with the display.

### DIFF
--- a/desmume/src/GPU.cpp
+++ b/desmume/src/GPU.cpp
@@ -475,9 +475,11 @@ void gpu_savestate(EMUFILE &os)
 	const GPUEngineB *subEngine = GPU->GetEngineSub();
 	
 	//version
-	os.write_32LE(2);
+	os.write_32LE(3);
 	
-	os.fwrite((u8 *)dispInfo.masterCustomBuffer, GPU_FRAMEBUFFER_NATIVE_WIDTH * GPU_FRAMEBUFFER_NATIVE_HEIGHT * sizeof(u16) * 2);
+	int bufferSize = dispInfo.customWidth * dispInfo.customHeight * dispInfo.pixelBytes * 2;
+	os.write_32LE(bufferSize);
+	os.fwrite((u8 *)dispInfo.masterCustomBuffer, bufferSize);
 	
 	os.write_32LE(mainEngine->savedBG2X.value);
 	os.write_32LE(mainEngine->savedBG2Y.value);
@@ -515,9 +517,22 @@ bool gpu_loadstate(EMUFILE &is, int size)
 		if (is.read_32LE(version) != 1) return false;
 	}
 	
-	if (version > 2) return false;
+	if (version > 3) return false;
 	
-	is.fread((u8 *)dispInfo.masterCustomBuffer, GPU_FRAMEBUFFER_NATIVE_WIDTH * GPU_FRAMEBUFFER_NATIVE_HEIGHT * sizeof(u16) * 2);
+	int currentBufferSize = dispInfo.customWidth * dispInfo.customHeight * dispInfo.pixelBytes * 2;
+	int streamBufferSize;
+	if (version >= 3)
+		is.read_32LE(streamBufferSize);
+	else
+		streamBufferSize = GPU_FRAMEBUFFER_NATIVE_WIDTH * GPU_FRAMEBUFFER_NATIVE_HEIGHT * sizeof(u16) * 2;
+
+	if (currentBufferSize < streamBufferSize)
+	{
+		is.fread((u8 *)dispInfo.masterCustomBuffer, currentBufferSize);
+		is.fseek(streamBufferSize - currentBufferSize, SEEK_CUR);
+	}
+	else
+		is.fread((u8 *)dispInfo.masterCustomBuffer, streamBufferSize);
 	
 	if (version >= 1)
 	{
--- a/desmume/src/GPU.cpp
+++ b/desmume/src/GPU.cpp
@@ -475,7 +475,7 @@ void gpu_savestate(EMUFILE &os)
 	const GPUEngineB *subEngine = GPU->GetEngineSub();
 	
 	//version
-	os.write_32LE(1);
+	os.write_32LE(2);
 	
 	os.fwrite((u8 *)dispInfo.masterCustomBuffer, GPU_FRAMEBUFFER_NATIVE_WIDTH * GPU_FRAMEBUFFER_NATIVE_HEIGHT * sizeof(u16) * 2);
 	
@@ -487,6 +487,8 @@ void gpu_savestate(EMUFILE &os)
 	os.write_32LE(subEngine->savedBG2Y.value);
 	os.write_32LE(subEngine->savedBG3X.value);
 	os.write_32LE(subEngine->savedBG3Y.value);
+
+	GPU->savestate(os);
 }
 
 bool gpu_loadstate(EMUFILE &is, int size)
@@ -513,11 +515,11 @@ bool gpu_loadstate(EMUFILE &is, int size)
 		if (is.read_32LE(version) != 1) return false;
 	}
 	
-	if (version > 1) return false;
+	if (version > 2) return false;
 	
 	is.fread((u8 *)dispInfo.masterCustomBuffer, GPU_FRAMEBUFFER_NATIVE_WIDTH * GPU_FRAMEBUFFER_NATIVE_HEIGHT * sizeof(u16) * 2);
 	
-	if (version == 1)
+	if (version >= 1)
 	{
 		is.read_32LE(mainEngine->savedBG2X.value);
 		is.read_32LE(mainEngine->savedBG2Y.value);
@@ -532,6 +534,11 @@ bool gpu_loadstate(EMUFILE &is, int size)
 		//subEngine->refreshAffineStartRegs(-1,-1);
 	}
 	
+	if (version >= 2)
+		GPU->loadstate(is);
+	else
+		GPU->loadstateDefault();
+
 	mainEngine->ParseAllRegisters();
 	subEngine->ParseAllRegisters();
 	
@@ -8398,6 +8405,43 @@ void GPUSubsystem::ClearWithColor(const u16 colorBGRA5551)
 			break;
 	}
 }
+
+void GPUSubsystem::savestate(EMUFILE &f)
+{
+	f.write_32LE(1); // version
+	f.write_floatLE(_displayInfo.backlightIntensity[0]);
+	f.write_floatLE(_displayInfo.backlightIntensity[1]);
+	f.write_floatLE(_backlightIntensityTotal[0]);
+	f.write_floatLE(_backlightIntensityTotal[1]);
+}
+
+bool GPUSubsystem::loadstate(EMUFILE &f)
+{
+	u32 version;
+	if (f.read_32LE(version) != 1) return false;
+
+	if (version == 1)
+	{
+		f.read_floatLE(_displayInfo.backlightIntensity[0]);
+		f.read_floatLE(_displayInfo.backlightIntensity[1]);
+		f.read_floatLE(_backlightIntensityTotal[0]);
+		f.read_floatLE(_backlightIntensityTotal[1]);
+	}
+}
+void GPUSubsystem::loadstateDefault()
+{
+	// UpdateAverageBacklightIntensityTotal adds to _backlightIntensityTotal, and is called 263 times per frame.
+	// Of these, 71 calls are after _displayInfo.backlightIntensity is set.
+	// This emulates those calls as a way of guessing what the backlight values were in a savestate which doesn't contain that information.
+	_backlightIntensityTotal[0] = 0.0f;
+	_backlightIntensityTotal[1] = 0.0f;
+	UpdateAverageBacklightIntensityTotal();
+	_displayInfo.backlightIntensity[0] = _backlightIntensityTotal[0];
+	_displayInfo.backlightIntensity[1] = _backlightIntensityTotal[1];
+	_backlightIntensityTotal[0] *= 71;
+	_backlightIntensityTotal[1] *= 71;
+}
+
 
 void GPUEventHandlerDefault::DidFrameBegin(const size_t line, const bool isFrameSkipRequested, const size_t pageCount, u8 &selectedBufferIndexInOut)
 {
--- a/desmume/src/GPU.h
+++ b/desmume/src/GPU.h
@@ -1861,6 +1861,10 @@ public:
 	template<NDSColorFormat OUTPUTFORMAT> void RenderLine(const size_t l);
 	void UpdateAverageBacklightIntensityTotal();
 	void ClearWithColor(const u16 colorBGRA5551);
+
+	void savestate(EMUFILE &f);
+	bool loadstate(EMUFILE &f);
+	void loadstateDefault();
 };
 
 class GPUClientFetchObject
--- a/desmume/src/NDSSystem.cpp
+++ b/desmume/src/NDSSystem.cpp
@@ -1229,7 +1229,7 @@ struct Sequencer
 		if (!divider.load(is)) return false;
 		if (!sqrtunit.load(is)) return false;
 		if (!gxfifo.load(is)) return false;
-		if (!readslot1.load(is)) return false;
+		if (version >= 4) if (!readslot1.load(is)) return false;
 		if (version >= 1) if(!wifi.load(is)) return false;
 #define LOAD(I,X,Y) if(!I##_##X##_##Y .load(is)) return false;
 		LOAD(timer,0,0); LOAD(timer,0,1); LOAD(timer,0,2); LOAD(timer,0,3); 
@@ -1760,7 +1760,7 @@ static bool loadUserInput(EMUFILE &is, int version);
 void nds_savestate(EMUFILE &os)
 {
 	//version
-	os.write_32LE(3);
+	os.write_32LE(4);
 
 	sequencer.save(os);
 
@@ -1780,7 +1780,13 @@ bool nds_loadstate(EMUFILE &is, int size)
 	u32 version;
 	if (is.read_32LE(version) != 1) return false;
 
-	if (version > 3) return false;
+	if (version > 4) return false;
+	// hacky fix; commit 281268e added to the saved info but didn't update version
+	if (version == 3)
+	{
+		if (size == 497)
+			version = 4;
+	}
 
 	bool temp = true;
 	temp &= sequencer.load(is, version);
--- a/desmume/src/saves.cpp
+++ b/desmume/src/saves.cpp
@@ -1125,7 +1125,8 @@ static bool ReadStateChunks(EMUFILE &is, s32 totalsize)
 		if (!is.read_32LE(t))  { ret=false; break; }
 		if (t == 0xFFFFFFFF) break;
 		if (!is.read_32LE(size))  { ret=false; break; }
-		
+		u32 endPos = is.ftell() + size;
+
 		switch(t)
 		{
 			case 1: if(!ReadStateChunk(is,SF_ARM9,size)) ret=false; break;
@@ -1159,6 +1160,13 @@ static bool ReadStateChunks(EMUFILE &is, s32 totalsize)
 			default:
 				return false;
 		}
+		if (is.ftell() != endPos)
+		{
+			// Should we just go ahead and return false?
+			msgbox->warn("There was an error loading the savestate. Your game session is probably corrupt now.");
+			is.fseek(endPos, SEEK_SET);
+		}
+
 		if(!ret)
 			return false;
 	}


zeromus:
I'm not sure I like the format of the state depending on whether HD scaling or >15bpp are used. If I take a savestate on one version of desmume, I should be able to load it regardless of my current settings. This is also an enormous amount of data. Consequently, we've had to add a "include screenshots in savestate" option in bizhawk. I see the value for casual gaming, though. I think as long as we retain the capability (even if we surface it as a user option yet) to make the 'screenshot' optional, and correctly handle the case of mismatches (by blacking out all the data which would have been in the screenshot), it should be OK

SuuperW:
You can load a savestate while using different settings than it was created with just as well on this version as you could previously. In both cases, the display immediately after loading will be incorrect.

The display could be re-formatted and scaled upon loading, but this would still result in a display that is not quite what it would have been with the current settings. (e.g. you could have HD scaling at 4 but if the state was created with HD scaling = 1, you'd have a low-res display just upscaled) Better than the current implementation in most cases, but not perfect.

As for the size of the savestate file, I do not see this as a major concern, at least with compression enabled. I created one savestate with default settings and another with 24bpp, HD scaling at 11, and texture upscaling at x4. The difference in file size is 1.3 MB. I'm sure this would be larger in more chaotic scenes, but I doubt many people are playing with such high settings anyway.
1.3MB would, however, be a much bigger concern in Bizhawk with automatic savestate generation every few frames. This, in addition to mismatches in settings, I think is enough reason to make saving the display an option. If we do this, I think there should also be an option to automatically advance one frame after loading a state which has no (or a mismatched?) display buffer.

zeromus:
I'd rather see the screen blanked instead of garbage. And that's a cryptic option nobody is ever going to choose. We'd have to default it to ON, and I'm not sure I like an emulator doing that by default.

SuuperW:
I have made the requested changes.

zeromus:
If _backlightIntensityTotal isn't even used right now, just leave a note about it commented out in the savestate logic and maybe a note written near the declaration that it needs to be savestated if used. Even then it doesnt **need** savestating, like in the crucial determinism-breaking way. It just means that the screenshots displayed after loadstate would look wrong, and those are of questionable value/reliability anyway.

SuuperW:
Regarding _backlightIntensityTotal:
Always using UpdateAverageBacklightIntensityTotal() does not work in all cases. The game may update the backlight registers at the end of a frame, after _displayInfo.backlightIntensity is set. In this case (which is the case in the game I tested with), loading a state and using UpdateAverageBacklightIntensityTotal() would result in the backlight level being set to what it should be next frame.
(And no, saving this and having to update the version number anyway is not part of my reason for wanting to save an HD framebuffer.)

Regarding saving the framebuffer:
1. The user will already (prior to any commits in this PR) see garbage upon loading a save state with different 3D settings than it was saved with, so I don't see this as a reason to not save the entire framebuffer. I also do not like the idea of clearing the framebuffer when loading a save state, as it would not be clear why the display isn't properly showing, or even if you know why you couldn't tell what settings to use in order to see that frame's display.

2. Any computer capable of running with a high resolution setting will have no problem saving or loading an HD framebuffer. Are you worried about performance when people share save state files with each other, using different specs/settings?
2a. How much should we be worrying about what the file size would be _without_ compression? As I mentioned earlier, the actual file size of a save state with a high-res framebuffer is much, much less than the expected non-compressed size. (Also who the heck uses a network drive for stuff like this? I wouldn't think this should be a concern of ours.)

3. While I haven't done anything with this yet, there actually is another framebuffer which must be saved in order to properly display after loading a save state: GPUEngineA._3DFramebufferMain or _3DFramebuffer16. Without saving these, 3D scenes appear as black one frame after loading a save state, as this buffer is cleared at some point in the load state process (probably during NDS_Reset()).  (In version 0.9.9, the 3D portion of the display is simply not updated that frame, rather than showing as black.) Thus in some situations, we have display issues for two frames after loading a state, not just one.
3a. Maybe one of you knows: Why are there separate 3D framebuffers for 16-bit and >16-bit color? I don't know of any situation where they would need to both exist at the same time.

rogerman:
@SuuperW: My response to each of your statements.

1. Sure, the user will see visual garbage, which is obviously not ideal. However, this only occurs for a single frame, and so the visual garbage is easily mitigated. If you don't want to see the visual garbage, then run the framebuffer at native size at 15-bit color.

If you are concerned about how clearing the framebuffer doesn't provide enough user feedback for indicating that loading a save state has occurred, then there is no problem here. In the Windows port, the HUD will report that a save state has been loaded. In the Cocoa port, the status bar will report that a save state has been loaded. Therefore, I see this as enough user feedback to warrant clearing the framebuffer after loading a save state.

Finally, the user's settings should have no bearing on the format of the save state file. Rather, the information saved in the save state file must be parsed properly to conform to the user's settings. It shouldn't matter what the user's settings are when the save state file is saved, nor should the save state file reflect what the user's settings were, since it is assumed that the save state file could be used on different host machines with different settings.

Look, if you really are bent on having a decently-looking framebuffer for that one frame after loading, then do this:
- Before you save the save state file, use a temporary in-memory buffer to resize the custom framebuffer down to 256x384 pixels and convert it to 16-bit 5551 colorspace (if needed), and then save that temp buffer to the save state file.
- As you load the save state file, read the framebuffer information to a temporary in-memory buffer that is 256x384 pixels for 16-bit colorspace. After you load the save state file, convert the temporary buffer to the user's colorspace (if necessary), and then resize the temporary buffer to fit the custom framebuffer.

Why do things this way? First, you preserve the layout of the save state file, so compatibility and performance are also preserved. Second, you get to fix the appearance of the output framebuffer (in a low-res way), regardless of the user's settings, which is another compatibility win. Third, it is better in this case to consume more CPU cycles instead of consuming more I/O bandwidth because the CPU always has better performance consistency (and is most likely faster) than I/O.

2. You failed to define what you mean by "no problem". If you mean that "the amount of data that is saved/loaded from file should not affect the ability to simply save/load the file on any machine" then, duh, there should be no problem. The "problem" I'm describing is that performance can be adversely affected when the user is running extremely large framebuffers, which is very possible on machines with fast CPUs.

When it comes to users sharing large save state files with each other: That's not even a concern in the slightest. The performance concern is for someone who wants to both run a large framebuffer and also heavily employ save states in their gameplay. For example, perhaps a TAS'er wants to play the game with consistent performance while wanting the graphics to look a little bit high-res, and so they run a modest HD resolution, like say 4x native 24-bit. Even then, the increased framebuffer size can cause a performance loss that could cause unwanted stuttering when no stuttering existed beforehand, which would make me howl in frustration if I were a TAS'er. Or maybe you can tell users that if they don't want stuttering of their save states, then they should upgrade to using dual high-performance NVMe SSDs in RAID-0. Right...

2a. First off, you cannot rely on compression as a performance example because our performance discussion deals with _data size_, not _file size_. Save state performance is not dependent on file size -- it is dependent on data size. It doesn't matter if we're using a perfect on-write compression algorithm that consumes zero CPU cycles because we still have deal with processing the actual data itself on the CPU and memory. Increasing the data size is never free! But even if increasing the data size was completely free for all computing resources, then what do we do? Tell users to use compression because @SuuperW wanted HD framebuffers in save state files? And if user's don't use compression, then they should just take the performance hit? I can foresee this change as being very unpopular with most users.

In addition, assuming that the user is even using compression in the first place, the myriad of different compression algorithms available also means that we don't know if compression will be performed on-write or post-write. For on-write compression, your machine will consume CPU cycles, which can cause other CPU-related performance issues. If we're talking about post-write compression, then the uncompressed save state file will still be pushed through I/O, so then we're back to our initial data size problem. Let's say that you're using Windows NTFS Full-Drive Compression, which is an on-write compression algorithm. Not only do most users leave this option off, but I can guarantee you that it does slow down drive writes, which will cause a noticeable stutter when saving a save state file.

As for the network drive issue: I was providing a concrete example of how a machine with a fast CPU can potentially be held back by slow I/O. Having a 6700K Intel computer running off a network drive was a clear example just to illustrate my point.

While you may not personally use a network drive, I know quite a few people who do. I've even set up a couple of offices to use network drives for their users. It's not an uncommon practice. But forget the network drive! What if the user is using a mechanical hard drive? Should save state loading/saving be slowed down for those users as well?

3. Admittedly, that is a weakness in the save state system. For SoftRasterizer, this can be fixed by reading the native-sized framebuffer from the save state and resizing it (and converting colorspace if necessary) into GPUEngineA::_3DFramebufferMain and GPUEngineA::_3DFramebuffer16. However, OpenGL is the odd one. It is possible to fix the 16-bit 3D framebuffer the same way as SoftRasterizer, but the main 3D framebuffer from OpenGL is actually a pointer to pinned GPU memory (done for performance reasons). It would be possible to fix this by uploading the save state's 3D framebuffer to the GPU, but uploading such a large block of data would cause a tremendously long stutter. That's why I decided not to touch this stuff for now.

3b. The main 3D framebuffer represents the 3D framebuffer slated to be used directly for display, and so it can be in whatever colorspace format it wants. This allows us to display the NDS video in 15-bit RGB555, 18-bit RGB666, or 24-bit RGB888. Therefore, we need this rendering target to support all of these colorspaces.

However, the 16-bit 3D framebuffer is a special rendering target that is guaranteed to be in the 16-bit colorspace. This is necessary for emulating NDS display captures, which only works in the 16-bit RGBA5551 colorspace. By having this special rendering target, we can fulfill this emulation requirement while maintaining the richer colorspace 18-bit RGB666 or 24-bit RGB888 at the same time.

SuuperW:
OK, I think I'm done with this now. All my tests with saving/loading states had the state's frame and the next frame displaying properly, using SoftRasterizer and OpenGL, native and upscaled resolution, and 15-bit and 24-bit color.

My sleep schedule has been completely wrecked this past week, so I may have missed something. Please review my changes and point out any potential problems you see. I've tried to be thorough.

rogerman:
The only issue I see is is where you unnecessarily changed the save state file layout again by moving the 3D framebuffer data from the GFX3D section to the GPU section. You didn't have to do that. Rather, it would have been better if you just stuck with the existing 3D framebuffer that was already being saved in GFX3D.

In commit 92e09f0, while it is true that the GFX3D framebuffer wasn't being used properly right now, that doesn't mean that it was ever NULL. The gfx3d_Update3DFramebuffers() was the hacky band-aid fix that prevented the save state file from reading/writing to NULL after I did a huge code refactor. In addition, it is a buffer that has historically been allocated there all along, so you might as well use it while preserving compatibility at the same time. (I know its confusing, but the whole save state file code is very wonky and needs to be completely reworked from the ground up.)

Note that the only difference between the main 3D framebuffer and the 16-bit 3D framebuffer is that the 16-bit 3D framebuffer guarantees an RGBA5551 colorspace while the other framebuffer has no colorspace guarantee. Therefore, when saving the save state file, you could have saved the main 3D framebuffer to GFX3D with a standardized RGBA8888 format (scaling down if necessary), and then when loading the save state file, reading back the RGBA8888 formatted framebuffer and converting it for both the main 3D framebuffer and the 16-bit 3D framebuffer.

Now the difficult part is making the main 3D framebuffer meaningful to OpenGL. When reading back this data from the save state file, you cannot simply write to the pointer returned by GPUEngineA::Get3DFramebufferMain() because OpenGL returns a pointer to READ-ONLY pinned GPU memory -- writing to this pointer directly is undefined behavior that could cause the GPU driver to crash. Instead, you would have to use OpenGL API to properly upload the data. I don't know if it's possible to use glBufferSubData() directly on the GL_PIXEL_PACK_BUFFER, or if you need to go as far as determining the final color output texture and uploading the data into that. But in either case, it is advantageous to upload only a native-sized 3D framebuffer since the total size would only be 192 KB. This data size is small enough to not cause any significant stutter on save state loading.

Also...

In commit 94d7340, where you're doing all the scaling/conversion stuff, there were existing functions to make things easier for you, and you would have gained optimizations as well.

- For colorspace conversions: Don't convert individual pixels -- use the ColorspaceConvertBuffer*() set of functions to convert the entire buffer at once.
- For scaling: We have the functions CopyLineExpand() and CopyLineReduce() in GPU.cpp. But admittedly, these functions were only meant for use in the GPU code alone, but now it looks like we'll need to use them for something more. In the future, I will need to make a proper API for these functions so that other programmers can access them more easily. It's also an excuse for me to clean up GPU.cpp some more.

@SuuperW: Though not perfect, I can clearly see your intentions in your code now. If you want to continue pursuing this, that's fine, because I have a lot of other issues on my plate at the moment. But if you want to be done with this, then I'll just approve of the changes as-is, and then I can take it from there. It looks like I'll need to make new changes to DeSmuME's API anyways.

zeromus:
I havent been paying attention, by why in the world do you need to scale down the display when saving a state? There should already be a 1x buffer. Scaling up when loadstating makes sense at least. The GPU interfaces should probably have a mechanism to import a 1x buffer and keep all the details to itself. 

We shouldnt need to save the 3d buffer. We should be able to regenerate it from cached geometry. That's the right way to do it. If we add this hack, then someone's less likely to do it the right way. If we commit this we at least need a common by the 3d buffer saving saying "THIS IS THE WRONG WAY. THE RIGHT WAY IS TO REGENERATE IT FROM THE CACHED GEOMETRY"

rogerman:
If the user is running HD, then you need to either save the entire HD framebuffer as-is (thereby changing the file layout) or scale it down to 1x (thereby keeping the layout the same). The latest discussion has brought us to rejecting saving the entire HD framebuffer in favor of scaling the HD framebuffer down to 1x precisely because the 1x buffer already exists.

As for the 3D framebuffer, that's just one way to handle this: By regenerating the 3D framebuffers through re-rendering. However, historically the 3D framebuffer has always existed in the save state layout, and so I wanted to give the rescaling/converting method a chance because it might ultimately result in a performance win (definitely on SoftRasterizer and possibly on OpenGL). That's why I didn't bother to mention the re-render method.

I personally don't care how a single frame might look after loading the save state because everything is going to be redrawn in the following frame. But I guess the 3D framebuffer might be off for a bunch of frames if the 3D doesn't render often enough (but the vast majority of games render at 30 FPS or more), and it would only be off because it won't match the user's rendering settings. I don't see that as too much of a problem.

zeromus:
Why does it matter if the 1x buffer already exists? Not wanting to put an entire HD framebuffer on disk makes sense, but worrying about what's already there doesn't make sense. Anyway I suppose you're trying to say when HD rendering is used there's no up to date 1x buffer to read from (I thought there was for capturing or something but maybe its not the whole 2 screen buffer) so the only way to make a 1x buffer in the first place is downscaling. if that's the case, that's fine. 

saving the 3d framebuffer won't result in a performance win. rendering a 3d frame to complete a loadstate is no more load than rendering a 3d frame any other time. It's not likely to exceed the cost of moving/compressing a similarly sized buffer by any significant margin. We shouldnt be saving data that can be regenerated from cached data, when that's an integral part of the hardware design. The framebuffer is a different story, as the hardware doesn't have the required elements to reconstruct the framebuffer (as it was based on timing).

SuuperW:
1. The 3D framebuffer was not being saved in the GFX3D section. As mentioned in my commit message for 92e09f0, the pointer _in the SF_GFX3D structure array_ was NULL. That entry in the array was being initialized at start-up with the _value_ of _gfx3d_colorMain, which is a pointer initialized to NULL.
I could put the 3D framebuffer save state logic in the gfx3d_savestate method if you prefer, but I don't see a good way to save it in the SF_GFX3D structure; the code for handling the SF structures isn't designed to work with data that moves around in memory. I suppose I could have the gfx3d_Update3DFramebuffers method update the SF structure, but that seems like a hacky fix which I'd like to avoid, especially since the entire SF structure setup seems ill-designed to begin with.

2. You say I could save the main 3D framebuffer with RGBA8888 color. But the main 3D buffer, as you said, does not guarantee that colorspace. 16-bit color was the only one that seems to be guaranteed anywhere, so I used that. (Plus, I kinda feel like the two saved framebuffers should be in the same format.)

3. For writing to the OpenGL buffer, the code as I have written it works in my machine, at least in the tests I did. It can write to the buffer returned by CurrentRenderer->GetFramebuffer(). However if this actually is undefined behavior and isn't guaranteed to work, I will look into using the OpenGL API.

4. I did a little looking for existing functions, but didn't know where to look and didn't find them. I'll definitely look into using those.

5. As for loading the 3D framebuffer versus regenerating it, I expect that loading it is _vastly_ faster than regenerating it. When I set 3D rendering to HD resolutions, my FPS drops drastically, indicating that the 3D rendering takes a significant amount of time. However when I was saving/loading HD framebuffers, there was almost no impact on save state load times versus when only loading a native-res framebuffer.

zeromus:
Guys, you dont need to render the 3d at the moment the state is loaded. You simply need to flag it for the next time a frame runs, in case the game doesn't submit new geometry that frame. Well, given the mess in the geometry engine right now, "simply" may be an exaggeration, but still.

SuuperW:
Then we'd often have to render the 3D twice in that frame, making the first frame after a save state take even longer than it already does. (It already takes a while because of the JIT.) I do not like this solution either.
(Note that the 3D framebuffer contains the 3D scene from the _previous_ frame. 3D scenes are shown displayed with a 1 frame diaplay lag. This is the entire reason that anything needs to be done at all.)

zeromus:
thats bogus. think about it more. at the instant we render 3d, we render whatever geometry we're supposed to render. as an optimization, we don't render it if it isn't "dirty". we would simply mark it as dirty when we load it from a state. 3d is rendered no more than once a frame in any case.

zeromus:
oh ok, youre right. what a pain in the ass. you see why this has been bugged for years? 

Actually, we need to save the 1x 3d buffer for determinism's purposes; since the game can read it back (by reading what's captured as it's rendered and mixed with 2d), this data cannot change between a savestate and later loadstate. If you want to save the HD 3d buffer too, that's up to you.

rogerman:
@SuuperW, a few more points:

1. I dispute the fact that _gfx3d_colorMain is ever NULL. On startup, when the GPUEngineA constructor runs, _gfx3d_colorMain is properly initialized. gfx3d_Update3DFramebuffers(), which is what updates _gfx3d_colorMain, is always immediately called whenever _gfx3d_colorMain could possibly point to a different location in memory. Stepping through with the debugger, I can verify that _gfx3d_colorMain is updating correctly at every instance. Your analysis is mistaken.

2. I should clarify my statement. When I said that the main 3D framebuffer does not guarantee a colorspace, that was only for when the framebuffer is initially generated, since it is possible that the 3D renderer could output either an RGBA6665 or RGBA8888 framebuffer.

However, for the save state file, you CAN guarantee the colorspace format by using colorspace conversion to force it into a specific format. Since RGBA8888 is the richer color format, this is the one I would recommend using when saving the main 3D framebuffer to the save state file. Then when you want to derive the 16-bit 3D framebuffer, simply convert the main 3D framebuffer from file (which should be in RGBA8888 format) to RGBA5551. That way, you only need to save a single 3D framebuffer. And guess what? The SF_GFX3D struct already reserves such a buffer for the save state file. Imagine that?

3. See https://www.khronos.org/registry/OpenGL-Refpages/gl2.1/xhtml/glMapBuffer.xml for more information. Of note, "_If a mapped data store is accessed in a way inconsistent with the specified access policy, no error is generated, but performance may be negatively impacted and system errors, including program termination, may result. Unlike the usage parameter of glBufferData, access is not a hint, and does in fact constrain the usage of the mapped data store on some GL implementations._"

4. All of the colorspace conversion functions are in colorspacehandler.cpp. But as for scaling... again... we don't have a nice API for that yet. The functions you're looking for are in GPU.cpp named CopyLineExpand() and CopyLineReduce(). Yeah, I should get to making a real scaling API at some point to make this kind of thing easier.

5. Loading the 3D framebuffers is undeniably faster for SoftRasterizer, and most likely faster for OpenGL. While you are correct here, your performance analysis is incorrect here. You cannot properly measure the performance of a one-shot operation by comparing it to a repeating operation.

The reason why loading the 3D framebuffers is faster than regenerating them via a re-render is because of instruction count and memory access. Besides running a test program for this, you can also look at the code disassembly for all the algorithms in question and make an estimate based on the number of instructions and memory accesses that you see.

When loading the 3D framebuffer, each source pixel's memory location is read exactly once and each destination pixel's memory location is written to exactly once. For each read/write operation, there are relatively few CPU instructions that go along with them, and so this can easily be seen as fast.

If you render with SoftRasterizer, then for each pixel in the target framebuffer, there are multiple read and writes to the target framebuffer itself, as well as multiple reads from several other memory locations, like vertices, indices, attributes, or texture memory. Of course, there are a huge number of CPU instructions that go along with the SoftRasterizer rendering algorithm, and so this must be slower than loading the 3D framebuffer.

If you render with OpenGL, then you don't have to worry so much about instruction count and memory accesses, but the PCI-E bus itself. There is always latency when transferring data across the bus, and downloading larger framebuffers from the GPU via glReadPixels() is always considered a slow operation. In DeSmuME's case, a user loading a save state would require downloading the 3D framebuffer at some point, but since save state loading occurs at line 0 of the GPU engine, this means that the GPU will need the results of the 3D framebuffer immediately. Therefore, you can't just flag the 3D framebuffer for it to finish downloading later -- the emulation timing doesn't allow for it! This is actually the worst-case scenario for glReadPixels(), because doing an asynchronous PBO transfer won't help hide latency since we need the 3D framebuffer immediately. And so because glReadPixels() is a bottleneck operation, it will be slower than simply loading the 3D framebuffer.

SuuperW:
@rogerman
1. _gfx3d_colorMain is NULL _before_ the constructor is called. Look at gfx3d.cpp in TASVideos/master, (https://github.com/TASVideos/desmume/blob/master/desmume/src/gfx3d.cpp) at line 286. It is set to NULL there. Then, at line 2681 _gfx3d_colorMain is accessed, still before the constructor, still NULL. You can put a breakpoint here to confirm. While you are correct that _gfx3d_colorMain gets properly set later and is properly updated, this value, the one copied into SF_GFX3D, is never updated.
Again, you can confirm this via debugging. Break at line 1078 in saves.cpp, then place a breakpoint at line 925 and continue to there.  Normally at this point in SubWrite(), the structure pointed to by the local variable sf will be zeroed out, as it would point to the last entry in the SFORMAT array. For the GFX3D array, however, it is the 2nd-to-last entry, with sf->desc being "G3CX".
And I take back what I said earlier about being able to fix this by updating the SF structure whenever _gfx3d_colorMain is updated; that buffer is not always native resolution, so we need code to handle that.

2. @zeromus does a native resolution 3D buffer always exist? The buffers I've been working with are scaled to match the user's resolution settings, and I do not see a 3D buffer which is always native resolution. If there isn't one already, then that's a separate determinism issue. (The scaling and color format set by the user's settings (as well as the user's choice of which renderer to use?) could slightly alter the contents of any buffer which is down-scaled or re-formatted to be a native resolution buffer to be read by the game.) And if there isn't:
@rogerman if doubling the size of the saved 3D buffer for the richer 888 color is worthwhile, then is it also a good idea to double the size of the saved display buffer to allow 888 color? And if yes, I don't see the logic in saying the buffer resolution should always be 1x. This doesn't seem consistent to me.

rogerman:
@SuuperW 

1. Oh, huh, you're right. Sorry about that. I had just recently worked around the requirement of SF structures needing statically allocated memory locations in WifiHandler, and so I still had that mindset while analyzing all of this.

Unfortunately, the quickest and easiest solution for this is to bring back the old gfx3d_convertedScreen buffer, which was a 256x192@32-bit statically allocated buffer that we used to use in GFX3D. A buffer like that would be used specifically for save state file purposes. While this does waste 192 KB of memory, I don't see a more expedient solution.

2. Wait... what? I never said "double the size of the saved 3D buffer" or something to that effect. What I'm saying is that the save state format in SF_GFX3D already reserves a 3D buffer of 256x192@32-bit for itself, so just use that -- no doubling of size required. But in order to standardize the data itself, it is better to save the buffer in RGBA8888 colorspace instead of RGBA6665 colorspace since RGBA8888 is the richer color format and takes up the exact same amount of space as RGBA6665.

And why save the display buffer in RGBA5551 format? For the exact same reason I've been making all along -- because the buffer is already reserved in the GPU save state format. Just use whatever is there so that you can avoid unnecessarily changing the save state file layout. There are also performance benefits to doing things this way as well.

SuuperW:
Are there plans to ever support a frame buffer scaled by a non-integer amount? There is some code to support this (see GPU.cpp's _gpuDstPitchCount), but it would make things simpler if I don't have to account for that.

rogerman:
The internal GPU code already does support non-integer scaling -- it is the DeSmuME clients that don't take advantage of it. And clients don't have to support non-integer scaling because it is completely optional. However, the internal GPU code needs to include it for the sake of completeness.

I don't see how things would be made "simpler if [you] don't have to account for [non-integer scaling]." To scale down a video line, just use CopyLineReduce() and it will take care of everything for you. You don't need to worry if the scaling is non-integer or not. Easy!

However, due to your PR still being on the table, I can't make any extensive changes to the API just yet, and so if you want to use CopyLineExpand() and CopyLineReduce() outside of GPU.cpp, then you will need to add the function prototypes to GPU.h. I will fix this later.

SuuperW:
@rogerman 
"if you want to be done with this, then I'll just approve of the changes as-is, and then I can take it from there."

I think this may be best. Sorry that I never got around to finishing what I wanted to do here.
Either way I'll almost certainly come back to do more work here eventually, but that probably won't be any time soon.

MikeKristof:
Alright, take care.Much Love

Mike Kristof



sob., 13 paź 2018 o 10:28 SuuperW <notifications@github.com> napisał(a):

PR #201, opened 2018-10-05
Fix undefined behaviour converting 3D postest registers

cbranch:
If the position test register is negative, conversion to unsigned
integer is undefined. This breaks games on arm64 where the behaviour is
defined as 'truncate to zero'. Converting to a signed integer first
guarantees the intended behaviour.


zeromus:
yuck. we need to get all the floating point out of here. but this will do for now.

PR #205, opened 2018-10-17
AntiAliasing Settings

Jules-A:
The automatic implementation is a major performance hog on low end machines, This patch allows people to LIMIT the amount of samples used to 8x, 4x and 2x while still using the in-place implementation for 8x and meaning it won't attempt to exceed or increase the framebuffer. The setting is disabled by default and is implemented in a way to satisfy previous concerns though I guess extra clarification could be provided.

FPS Testing:
> FPS: MSAA Off, Edge marking on = 67
> FPS: MSAA On, Default settings(auto=16), Edge marking on = 51  (~31% drop)
> FPS: MSAA On, Default settings(auto=16), Edge marking off = 54  (~24% drop)
> FPS: MSAA On, x8 Sample limit, Edge marking on = 54  (~24% drop)
> FPS: MSAA On, x8 Sample limit, Edge marking off = 56  (~20% drop)
> FPS: MSAA On, x4 Sample limit, Edge marking on = 61  (~10% drop)
> FPS: MSAA On, x4 Sample limit, Edge marking off = 64  (~5% drop)
> FPS: MSAA On, x2 Sample limit, Edge marking on = 64  (~5% drop)
> FPS: MSAA On, x2 Sample limit, Edge marking off = 67  (~0% drop)

Edge marking is needed with MSAA disabled but can be disabled in the games I've tested with MSAA to look better and increase FPS.

Comparisons: [MSAA Off](https://user-images.githubusercontent.com/1760158/47053628-8049cd80-d1e0-11e8-88ec-334317de9f9a.JPG),
 [MSAA 2x Edge marking off](https://user-images.githubusercontent.com/1760158/47053806-40cfb100-d1e1-11e8-98fc-d5428714130a.JPG),
[MSAA 16x Edge marking on](https://user-images.githubusercontent.com/1760158/47053736-ecc4cc80-d1e0-11e8-8f65-2ec1a3418b40.JPG) .

As you can see, with 2x on it's miles better than disabled while having 0 impact when disabling edge marking. 16x doesn't look that much better and I can hardly tell the difference between 4x while it made the difference between unplayable FPS or not.

I also made the menu look a little better imo (less wasted space): 
[Old menu](https://user-images.githubusercontent.com/1760158/47053935-e8e57a00-d1e1-11e8-9355-b3fc1d260d86.JPG)
[New menu](https://user-images.githubusercontent.com/1760158/47091000-9d20e800-d256-11e8-9560-a4c30ef071f5.JPG)

Other stuff: Changes are only for Windows, updated gitignore for vs2017 15.8.7, Nvidia and Intel currently untested, don't expect big gains on high end graphics. At higher (unplayable) settings I've seen up-to a 45% performance drop from auto-msaa. Tested on a HD 5750 1GB and FX8320.

### DIFF
--- a/desmume/src/frontend/windows/main.cpp
+++ b/desmume/src/frontend/windows/main.cpp
@@ -426,7 +426,7 @@ int sndvolume=100;
 
 const int possibleMSAA[] = {0, 2, 4, 8, 16, 32};
 int maxSamples = 0;
-bool tempContextAttempted = false;
+bool didGetMaxSamples = false;
 
 SoundInterface_struct *SNDCoreList[] = {
 	&SNDDummy,
@@ -5681,24 +5681,29 @@ LRESULT CALLBACK GFX3DSettingsDlgProc(HWND hw, UINT msg, WPARAM wp, LPARAM lp)
 	{
 	case WM_INITDIALOG:
 		{
-			int i, z = 0;
 			const char MSAADescriptions[6][9] = {"Disabled", "2x", "4x", "8x", "16x", "32x"};
 
-			if(cur3DCore != 0 && cur3DCore != 2) //Get max device samples from the current renderer only if it's OpenGL.
-				maxSamples = CurrentRenderer->GetDeviceInfo().maxSamples;
-			if (maxSamples == 0 && tempContextAttempted == false) 
-			{ //If it already has a value it's likely correct so don't create again.
-				tempContextAttempted = true;
-				bool isTempContextCreated = windows_opengl_init(); //Create a context just to get max device samples.
-				if(isTempContextCreated) //Creating it here because it's only needed in this window.
+			if (!didGetMaxSamples)
+			{
+				if (cur3DCore != 0 && cur3DCore != 2) //Get max device samples from the current renderer only if it's OpenGL.
 				{
-					GLint maxSamplesOGL = 0;
+					maxSamples = CurrentRenderer->GetDeviceInfo().maxSamples;
+					didGetMaxSamples = true;
+				}
+				else
+				{
+					bool isTempContextCreated = windows_opengl_init(); //Create a context just to get max device samples.
+					if (isTempContextCreated) //Creating it here because it's only needed in this window.
+					{
+						GLint maxSamplesOGL = 0;
 #if defined(GL_MAX_SAMPLES)
-					glGetIntegerv(GL_MAX_SAMPLES,&maxSamplesOGL);
+						glGetIntegerv(GL_MAX_SAMPLES, &maxSamplesOGL);
 #elif defined(GL_MAX_SAMPLES_EXT)
-					glGetIntegerv(GL_MAX_SAMPLES_EXT,&maxSamplesOGL);
+						glGetIntegerv(GL_MAX_SAMPLES_EXT, &maxSamplesOGL);
 #endif
-					maxSamples = maxSamplesOGL;
+						maxSamples = maxSamplesOGL;
+						didGetMaxSamples = true;
+					}
 				}
 			}
 
@@ -5720,8 +5725,8 @@ LRESULT CALLBACK GFX3DSettingsDlgProc(HWND hw, UINT msg, WPARAM wp, LPARAM lp)
 			CheckDlgButton(hw,IDC_TEX_DEPOSTERIZE, CommonSettings.GFX3D_Renderer_TextureDeposterize);
 			CheckDlgButton(hw,IDC_TEX_SMOOTH, CommonSettings.GFX3D_Renderer_TextureSmoothing);
 
-			SendDlgItemMessage(hw, IDC_NUD_PRESCALEHD, UDM_SETRANGE, 0, MAKELPARAM(16, 1));
-			SendDlgItemMessage(hw, IDC_NUD_PRESCALEHD, UDM_SETPOS, 0, video.prescaleHD);
+			SendDlgItemMessage(hw,IDC_NUD_PRESCALEHD, UDM_SETRANGE, 0, MAKELPARAM(16, 1));
+			SendDlgItemMessage(hw,IDC_NUD_PRESCALEHD, UDM_SETPOS, 0, video.prescaleHD);
 
 			CheckValidMSAA(CommonSettings.GFX3D_Renderer_MultisampleSize);
 			if (CommonSettings.GFX3D_Renderer_MultisampleSize > maxSamples) 
@@ -5730,15 +5735,18 @@ LRESULT CALLBACK GFX3DSettingsDlgProc(HWND hw, UINT msg, WPARAM wp, LPARAM lp)
 				WritePrivateProfileInt("3D", "MultisampleSize", maxSamples, IniName);
 			}
 
-			for (i = 0; z <= maxSamples; i++)
+			// Generate the MSAA pop-up menu
+			ComboBox_AddString(GetDlgItem(hw,IDC_MULTISAMPLE_SIZE), MSAADescriptions[0]);
+			ComboBox_SetCurSel(GetDlgItem(hw,IDC_MULTISAMPLE_SIZE), 0);
+			for (int i = 1, z = 2; z <= maxSamples; i++, z*=2)
 			{
 				ComboBox_AddString(GetDlgItem(hw,IDC_MULTISAMPLE_SIZE),MSAADescriptions[i]);
 				if (z == CommonSettings.GFX3D_Renderer_MultisampleSize)
-					ComboBox_SetCurSel(GetDlgItem(hw,IDC_MULTISAMPLE_SIZE),i);
-				if(z >= 2) z *= 2; else z=2; //Double the samplesize if 2 or more;
+					ComboBox_SetCurSel(GetDlgItem(hw,IDC_MULTISAMPLE_SIZE), i);
 			}
-
-			for(i = 0; core3DList[i] != NULL; i++)
+			
+			// Generate the 3D Rendering Engine pop-up menu
+			for (int i = 0; core3DList[i] != NULL; i++)
 			{
 				ComboBox_AddString(GetDlgItem(hw, IDC_3DCORE), core3DList[i]->name);
 			}


rogerman:
@Jules-A, thanks for your contribution, but I have some mixed feelings about the way that all of this was implemented. But before we get to that, I have some questions about your testing methodology:

1. What is the GPU Scaling Factor that you are testing with? (I'm assuming that you used the same GPU Scaling Factor throughout all your testing... I hope.)

2. What GPU hardware are you using? There are definite MSAA differences between, say, an Nvidia GT 730 vs an Intel Iris Plus 655, or an Nvidia GT GTX 1080 Ti vs an Intel GMA 4500. And as a related question -- if you are on a desktop PC, then are you are running your graphics card on a PCI-E 3.0 16x slot?

3. What game(s) did you do your testing on?

4. What scene(s) in the games did you do your testing on?

5. A 45% performance drop is an extraordinary scenario that demands extraordinary details. In my personal testing, which has involved many many many many GPUs running under many OS/hardware combinations, I have never experienced a performance drop this large. So again, in what game and in which scene did this performance drop occur?

Now then... in reviewing your code...

- It looks like your code simply sets user-specified MSAA limits that override the built-in limits, and you even made UI elements for each of these limits. So if you're going to go this far, then why not go all the way and allow the user to select their own MSAA value? That would be the better solution at this point.

- Your code does not error-check for invalid MSAA values. Sure, you personally set the Windows-specific code to hard-set the values, but you can't assume that this will always be the case. You should bulletproof your code by limiting the specific MSAA values to 1, 2, 4, 8, 16, and 32, regardless of what the user may try to set it to.

Jules-A:
> 
> 
> @Jules-A, 
>     1. What is the GPU Scaling Factor that you are testing with? (I'm assuming that you used the same GPU Scaling Factor throughout all your testing... I hope.)

3x, though I did test other factors as well, 2x offered slightly less of a fps boost while 5x (while not playable offered more). Obviously I kept it the same...
> 
>     2. What GPU hardware are you using? There are definite MSAA differences. Are you are running your graphics card on a PCI-E 3.0 16x slot?
Currently a HD 5750 1GB as I just sent my RX570 in for RMA (it actually had graphics corruption with MSAA disabled and was fixed by enabling it of all things though it also suffered a large FPS loss from MSAA). As I'm still running an FX8320, it is running on PCI-E 2.0 16x.
> 
>     3. What game(s) did you do your testing on?
FPS numbers are from Pokemon HG, though I tested Platinum and Black 1. I don't have copies of any other games to test unfortunately.
> 
>     4. What scene(s) in the games did you do your testing on?
The most demanding scene area that I had found so far in HG, a run that moves from Celadon city, loops Saffron and back to Celadon. I've used the same area and runs for multiple benchmarks over multiple Desmume builds.
> 
>     5. A 45% performance drop is an extraordinary scenario that demands extraordinary details. In my personal testing, which has involved many many many many GPUs running under many OS/hardware combinations, I have never experienced a performance drop this large. So again, in what game and in which scene did this performance drop occur?

Yes, it was testing with x5 GPU scaling factor while having another game open, though FPS was terrible obviously so it's a worst case scenario.
> 
> 
> Now then... in reviewing your code...
> 
>     * It looks like your code simply sets user-specified MSAA limits that override the built-in limits, and you even made UI elements for each of these limits. So if you're going to go this far, then why not go all the way and allow the user to select their own MSAA value? That would be the better solution at this point.
> 
Sure, it would be better but I wanted to make it so it was like the user was choosing to override the default settings to address concerns of people accidentally setting it. Maybe it could be more noticeable but I'm not sure what to do in that area other than waste space on a message or give a message box. There is also pretty much no point in choosing x16 as auto pretty much does that already and when you get to higher resolutions, x8 can get pretty pointless even (32 is just stupid). That is why I only gave 2x, 4x, 8x. 
Also I'm not sure how I'd go about increasing the framebuffer so it can run at say x16 when over 4x scaling factor anyways and if there would be any graphics benefit at all.
>     * Your code does not error-check for invalid MSAA values. Sure, you personally set the Windows-specific code to hard-set the values, but you can't assume that this will always be the case. You should bulletproof your code by limiting the specific MSAA values to 1, 2, 4, 8, 16, and 32, regardless of what the user may try to set it to.

Oh wow, how did I manage to miss that? Thanks.

rogerman:
It doesn't look like your testing methodology is very good here, based on all the information you've presented here.

MSAA performance testing is very dependent on memory bandwidth performance, so you should be focusing your testing configuration on that. That means you should NOT be using the following settings for your MSAA testing: Texture Scaling, Texture Deposterize, Edge Mark, or Fog. But to make it easier to differentiate certain smoothing effects, you should turn on Smooth Textures. The settings that you are using in your 3D Settings window do not reflect this configuration.

Edge Mark gets a special mention here: Due to the way the Edge Mark algorithm works, it can really be thrown off when MSAA is used. Therefore, you should NEVER use Edge Mark and MSAA together due to the graphical glitches that will result. Any MSAA test that is run with Edge Mark enabled is a meaningless test.

In order to reduce the performance influence of non-GPU related stuff, you should also minimize your CPU usage as much as possible. This means that you should be running a 15-bit Color Depth, turning off Advanced Bus-Level Timing, and running Dynamic Recompiler with a block size of 100. For SPU settings, you should use Dual SPU Synch/Asynch and turn off Advanced SPU Logic, as well as changing the Sound Core to None.

I'll mention this just in case: Client-side graphics enhancements, such as Magnification Filters, should be turned off.

Finally, certain game scenes can influence performance in ways not related to MSAA. It is very important to know what the 3D rendering FPS is (as opposed to the client's video FPS) and to compare that to the ARM9 CPU load average. Ideally, you want a game scene that uses a lot of polygons while running at 60 FPS with a very low ARM9 CPU load average.

You also want to use games that don't make use of the internal NDS frame capture, which all dual-screen 3D games will use. In other words, stick with games that only use one screen for 3D.

Some good game scenes for testing MSAA performance are:
- "Kingdom Hearts: Re-coded": Inside the Destiny Islands Secret Place cave. (30 FPS, 16% ARM9)
- "Super Mario 64 DS": Inside Princess Peach's room. (30 FPS, 12% ARM9)
- "Solatorobo: Red the Hunter": East Shipyard, at the top of the first area of Airedale. (60FPS, 30% ARM9)
- "Dementium: The Ward": In the initial room the player starts in. This is arguably the best scene for this kind of testing. (60 FPS, 9% ARM9)

For OpenGL in particular, certain game scenes with a large number of polygons of dissimilar properties will greatly hurt OpenGL performance in a way that negatively affects MSAA performance without MSAA actually being at fault. These games and/or game scenes should be avoided in MSAA testing. OpenGL really hates doing state changes and will punish you with slow performance all around if you do state changes a lot. You can tell if this scenario is occurring if you find yourself getting greatly improved performance when switching to SoftRasterizer. (And in case your wondering -- NO, we can't just optimize this and make it all "go faster" because we need to emulate how the NDS is doing the polygon state changes, and it just so happens that OpenGL doesn't like doing it. Blame OpenGL for always wanting to do slow state validation.)

Let's look at some actual numbers. All testing was performed on a 2008 MacBook Pro with a 2.53 GHz Intel Core2Duo T9400 and Nvidia GeForce GT 9600M on a PCI-E 2.0 16x bus (note that I'm purposefully using hardware that is close, but slightly inferior, to yours in this testing) using the following parameters:
- 15-bit GPU Color Depth
- Texture Scaling Factor = 1
- Texture Deposterize = OFF
- Texture Smoothing = ON
- Edge Mark = OFF (this is very important!)
- Fog = OFF
- Advanced Bus-Level Timing = OFF
- Dynamic Recompiler = ON, with Block Size = 100
- Sound Synchronization = Dual SPU Synch/Asynch
- Advanced SPU Logic = OFF
- Audio Output = None
- The DeSmuME version used is the Cocoa port based off of commit 01225c65.

I chose this combination of CPU and GPU because the T9400 is fairly matched in performance to the 9600M. Also, the 9600M is the closest performance match to your Radeon HD 5750 of the hardware I've got, and both GPUs have a GL_MAX_SAMPLE_SIZE of 8 (although the 9600M can achieve a GL_MAX_SAMPLE_SIZE of 16 on Windows, but I'm testing DeSmuME on macOS in order to force the sample sizes to match).

All testing was run using the OpenGL renderer unless otherwise noted.

**Kingdom Hearts, GPU Scaling Factor 1, No MSAA**
image

**Kingdom Hearts, GPU Scaling Factor 1, With MSAAx8 (No performance difference)**
image

**Kingdom Hearts, GPU Scaling Factor 3, No MSAA**
image

**Kingdom Hearts, GPU Scaling Factor 3, With MSAAx8 (~11% slower than No MSAA)**
image

**Super Mario DS, GPU Scaling Factor 1, No MSAA**
image

**Super Mario DS, GPU Scaling Factor 1, With MSAAx8 (no performance difference)**
image

**Super Mario DS, GPU Scaling Factor 3, No MSAA**
image

**Super Mario DS, GPU Scaling Factor 3, With MSAAx8 (~10% slower than No MSAA)**
image

**Solatorobo, GPU Scaling Factor 1, No MSAA**
image

**Solatorobo, GPU Scaling Factor 1, With MSAAx8 (no performance difference)**
image

**Solatorobo, GPU Scaling Factor 3, No MSAA**
image

**Solatorobo, GPU Scaling Factor 3, With MSAAx8 (~6% slower than No MSAA)**
image

**Dementium, GPU Scaling Factor 1, No MSAA**
image

**Dementium, GPU Scaling Factor 1, With MSAAx8 (no performance difference)**
image

**Dementium, GPU Scaling Factor 3, No MSAA**
image

**Dementium, GPU Scaling Factor 3, With MSAAx8 (no performance difference)**
image


This all leads us to...

> FPS numbers are from Pokemon HG, though I tested Platinum and Black 1. I don't have copies of any other games to test unfortunately.

And that's your problem right there. **The Pokemon games, such as HeartGold or Black/White, are very poor indicators for MSAA performance testing.**

For example, in HeartGold in New Bark Town (the starting town for the player), I'm getting a 53% ARM9 load average for a game that only runs at 30 FPS. Clearly, the ARM9 emulation is being overly influential on our test results. In fact, this is basically true for all the outdoor areas.

**Pokemon HeartGold, GPU Scaling Factor 1, OpenGL No MSAA**
image

**Pokemon HeartGold, GPU Scaling Factor 1, OpenGL With MSAAx8 (no performance difference)**
image

**Pokemon HeartGold, GPU Scaling Factor 3, OpenGL No MSAA**
image

**Pokemon HeartGold, GPU Scaling Factor 3, OpenGL With MSAAx8 (~32% slower than No MSAA)**
image

**Pokemon HeartGold, GPU Scaling Factor 1, SoftRasterizer No MSAA (~72% faster than OpenGL 1x)**
image

**Pokemon HeartGold, GPU Scaling Factor 3, SoftRasterizer No MSAA (~47% faster than OpenGL 3x)**
image


And just for kicks, let's do some testing at GPU Scaling Factor 5x.

**Pokemon HeartGold, GPU Scaling Factor 5, OpenGL No MSAA**
image

**Pokemon HeartGold, GPU Scaling Factor 5, OpenGL With MSAAx8 (~39% slower than No MSAA)**
image

**Pokemon HeartGold, GPU Scaling Factor 5, SoftRasterizer No MSAA (~27% slower than OpenGL 5x)**
image

The fact that SoftRasterizer 3x is 47% faster than OpenGL 3x on a T9400 Core2Duo vs. GeForce GT 9600M proves that Pokemon HeartGold is getting severely affected by OpenGL's state validation slowness, at least in this outdoor scene. Also, notice how at 5x, SoftRasterizer is only slower than OpenGL by 27%, where this difference will be much higher for other games. Again, this is OpenGL state validation affecting MSAA performance, which is another reason that makes Pokemon HeartGold a bad MSAA performance indicator.

I have a sneaking suspicion that your Hotel Grand Lake scene is also forcing OpenGL to do a lot of state changes. I'm betting that if you switch to SoftRasterizer, then it will speed up tremendously. This is probably the worst possible scene for MSAA testing due to the many outside influences on MSAA's performance. But hey, I'm just guessing here and I could be wrong. Could you upload the .dsv file with you near the Hotel Grand Lake scene so that I could confirm your performance drop?

In conclusion: My testing, using hardware that is close, but slightly inferior, to yours shows that the automatic MSAA sample size selection does not pose any significant performance problems for low-end GPU hardware. Both the Radeon HD 5750 and the GeForce GT 9600M on macOS use MSAAx8 for 1x and 3x GPU Scaling Factors, and running MSAAx8 for both sizes yields a performance penalty of 11% or less. Testing OpenGL 3x MSAAx8 on Pokemon HeartGold is an outlier with a performance penalty of 32%, but this can be explained by the high ARM9 load average of 53% and the excessive polygon state changes that both indirectly hurt MSAA performance in OpenGL.

**In short, I personally don't see any problems here, but you do see a problem here. So for your sake, let's try things your way and try to do something about this to help MSAA performance for outlier games running on low-end GPUs.**

> Sure, it would be better but I wanted to make it so it was like the user was choosing to override the default settings to address concerns of people accidentally setting it.

I don't see how anyone could "accidentally" set anything. The user would be explicitly choosing which MSAA setting they want (via a checkbox, menu options, or whatever), and then they would have to confirm it by clicking OK to exit the 3D Settings window. Nothing accidental here.

> Maybe it could be more noticeable but I'm not sure what to do in that area other than waste space on a message or give a message box.

There would be no waste of space involved. If you're going to reject the automatic selection of the MSAA sampling size, I'm suggesting that you simply go all the way. **You should allow the user to choose whatever MSAA sampling size they want.** You can allow the choice through radio buttons, a drop-down menu, a text field with a pair of up-down arrows -- there are several UI choices available to you that would use only minimal space. No message box would be needed here -- just have a label next to the control(s) that says "MSAA Samples" or something like that.

> There is also pretty much no point in choosing x16 as auto pretty much does that already and when you get to higher resolutions, x8 can get pretty pointless even (32 is just stupid).

This is purely a subjective opinion on your part. You may think that x8 is sufficient but nearly pointless, x16 is totally pointless, and x32 is "just stupid". Well I can assure you that many people would vehemently disagree with you. In fact, the reason why the automatic MSAA selection goes up to x32 is because of an impassioned defense for this mode by one of our users (I originally only wanted it to use x16 as the maximum MSAA). Since x32 is only intended to work at 1x GPU scaling, I didn't see this as a problem. But you, with your Radeon HD 5750 running on a PCI-E 2.0 bus, do think that this is a problem. **And so the best choice at this point is to just have explicitly user-selectable MSAA sample sizes.**

> That is why I only gave 2x, 4x, 8x. If you or anyone has suggestions on implementation it would be awesome.

**Again, user-selectable MSAA sample sizes.** I'm pretty sure that this is what you really want -- not this hybrid mode of "limiting-the-MSAA-sample-size-of-the-automatic-selector" business.

> Also I'm not sure how I'd go about increasing the framebuffer so it can run at say x16 when over 4x scaling factor anyways and if there would be any graphics benefit at all.

You shouldn't need to worry about those details because OpenGLRenderer::CreateMultisampledFBO() should already take care of that for you. The only thing that you are responsible for is ensuring that OpenGLRenderer::CreateMultisampledFBO() receives a valid MSAA sample size. Generating a valid MSAA sample size is the client's responsibility. In other words, as long as you don't use an invalid MSAA sample size, then everything should just work.

And is there a graphical benefit? I am assured by users that there is certainly a graphical benefit here, and I am personally inclined to agree with them. However, is the graphical benefit of going from MSAAx8 to MSAAx16 worth the performance cost? You don't think so. **Which again, is a reason why you should implement user-selectable MSAA sample sizes so that everyone can be happy here.**

Jules-A:
I appreciate the time you took to benchmark but I really don't see the point to just test MSAA performance alone, but instead at values people would choose. Sure it's not going to cause fps drops when it's not memory limited but the point of limiting MSAA in the first place is to help lower the bottleneck while sacrificing less visual quality than other settings. Also, AMD's OpenGL experience under Windows really isn't very good, it may also be contributing to the issue.

>Again, user-selectable MSAA sample sizes. I'm pretty sure that this is what you really want -- not this hybrid mode of "limiting-the-MSAA-sample-size-of-the-automatic-selector" business.

Actually, all I ever wanted to do was allow lowering MSAA samples and leave the rest to the auto settings but I was still going to implement a drop-down. If you feel that's what should be done, that's what I shall do.

> Therefore, you should NEVER use Edge Mark and MSAA together due to the graphical glitches that will result.

Shouldn't it be automatically disabled then when MSAA is enabled? A normal user wouldn't have any idea about that. I always thought Edge Mark was required by some games to prevent graphical glitches.

>The Pokemon games, such as HeartGold or Black/White, are very poor indicators for MSAA performance testing.

Actually Black 1 in the areas I've tested (I haven't left the starting area) did not experience as large of a drop as Plat or HG. I haven't played any other games so obviously I didn't know that. Regardless I'd argue the Pokemon games are one of the most emulated so you can't just rule them out as an anomaly.

>Could you upload the .dsv file with you near the Hotel Grand Lake scene so that I could confirm your performance drop?

[Pokemon - Platinum.zip](https://github.com/TASVideos/desmume/files/2490432/Pokemon.-.Platinum.zip)


You raised many points, I'll reply to more when I get more time.

Jules-A:
>You're saying this now? The very existence of this PR is completely based off of your own performance test!
I'm sorry I forgot to mention, I usually post late at night, I was actually going to re-do the benchmarks in the morning but saw you'd done extensive testing already on similar settings that I used with results that confirmed my findings (at least for the only game I recorded results for). 

>I'm not getting a 42% drop in this scene

If where you tested was where you took the screenshots then that was not the location, it was actually indoors exactly where the save was at. Performance is obviously better than outdoors so that area isn't really an issue but that's where I was able to reproduce such odd differentials.

>The printf statement that you added for reporting Max Multisamples was incorrectly done, making you then suggest that the automatic MSAA selector is broken. This claim is based on a wrongly reported MSAA sample size. Also, Texture Scaling Factor has no effect on the automatic selector. Rather, the automatic selector is based off of GPU Scaling Factor, in which the automatic selector should lower the MSAA sample size as the GPU Scaling Factor increases to certain thresholds. I can assure you that the automatic selector does work as intended.

It was placed in OpenGLRenderer::GetLimitedMultisampleSize() right before `if (deviceMultisamples > maxMultisamples)
	{
		deviceMultisamples = maxMultisamples;
	}`
So it was just getting set to maximum supported basically. This didn't happen when I changed GPU scaling (yes I got it mixed up in previous post) to 8/9/10 but disabling/renabling MSAA it just reported 32, along with on subsequent startups (with x10 GPU scale) whereas it would report the correct value when I used the limiter from this patch (it was meeting my conditions but defaulting to tier1 conditions without). 

>Now is it possible to make the mutual exclusivity of Edge Mark and MSAA an automatic behavior? Sure! But then the question arises: Which option should take priority? Should Edge Mark disable MSAA, or should MSAA disable Edge Mark? There will certainly be differing opinions amongst users about this. Therefore, I won't make this behavior automatic because it would require additional UI for users to choose whether Edge Mark or MSAA should take priority.

Another UI option wouldn't be needed, just another preference or 2 that would save individual values but you would probably want to give notice the first time it automatically disabled marking so the user knows. 
Lets just say we create 2 new bools, softMarking (default true) and oglMarking (default false). We simply need to check which renderer is being used/going to be used (incase they change both at same time) then use the corresponding variable. Then if they choose to enable/disable edge marking we just we just save the value to the corresponding variable based on renderer.

rogerman:
> It was placed in OpenGLRenderer::GetLimitedMultisampleSize() right before if (deviceMultisamples > maxMultisamples) { deviceMultisamples = maxMultisamples; }
So it was just getting set to maximum supported basically.

Okay, so there's your problem right there.

The method that actually sets everything up based on an MSAA size is CreateMultisampledFBO(), in which the MSAA size it uses is the one returned by GetLimitedMultisampleSize(). So what value is actually returned by GetLimitedMultisampleSize()? (Hint: It isn't going to be 32 on your Radeon HD 5750!) You are printf'ing the wrong value, and that's why this particular test of yours is bogus.

Sigh... but maybe I should get to fixing the minor bug I was referring to earlier, as that might be the source of some confusion on your part. (Even though it should have had zero impact on your actual performance tests.)

> **Another UI option wouldn't be needed**, just _another preference or 2_ that would save individual values but you would probably want to _give notice the first time it automatically disabled marking_ so the user knows.

You do realize that you just contradicted yourself? The user is going to have to set up these other "preference or 2" somehow, which would require additional UI. Giving "notice the first time so the user knows" also requires additional UI.

The current implementation of the mutual exclusivity of Edge Mark and MSAA being an automatic behavior still holds. We just trust the user to either do the right thing here, or to knowingly tolerate the graphical glitches of doing the wrong thing.

rogerman:
Alright, I fixed the bug with the automatic MSAA sample size selector in commit 68e91ec.

rogerman:
> Format:
Location (cpu)
MSAA FPS Off, On (difference)

>HG
celadon fountain (60%)
84, 65 (29%)
vermillion port near houses (70%)
68, 57 (19%)
enterence to bell tower (50%)
83, 66 (26%)
celadon mall elevator (45%)
105, 79 (33%)
PLAT
lake hotel (50%)
103, 78 (32%)
hearthhome city outside pokecenter (70%)
69, 56 (23%)
pokemon league outside entrance (50%)
75, 61 (23%)
BLACK
Starting area near first grass (37%)
88, 79 (11%)

Ahh... much more reasonable to expectations -- thanks for doing this for me. Your report of 42% performance reduction with MSAA alone at only 3x GPU Scaling Factor was downright scary. We're only talking about doing MSAAx8 on a 768x576 framebuffer, and so an older not-so-terrible GPU should be able to handle this kind of load without having a total performance catastrophe.

But now let's get to the real issue I'm having...

@Jules-A, truth be told, your PR is blocking some changes that I want to make, and so to speed things up, how about this -- I'll just implement user-selectable MSAA sample sizes at the backend level myself, and also add the UI for this feature in the Cocoa port. You will be responsible for implementing the UI for the Windows port. How's that sound? I have some fairly extensive changes that I want to commit really soon, but I really need you to finish up with NDSSystem.h and frontend/windows/main.cpp first as to avoid some merging grief.

rogerman:
>Format:
Location (cpu)
MSAA FPS Off, On (difference)

>HG
celadon fountain (60%)
84, 65 (29%)
vermillion port near houses (70%)
68, 57 (19%)
enterence to bell tower (50%)
83, 66 (26%)
celadon mall elevator (45%)
105, 79 (33%)
PLAT
lake hotel (50%)
103, 78 (32%)
hearthhome city outside pokecenter (70%)
69, 56 (23%)
pokemon league outside entrance (50%)
75, 61 (23%)
BLACK
Starting area near first grass (37%)
88, 79 (11%)
Ahh... much more reasonable to expectations -- thanks for doing this for me. Your report of 42% performance reduction with MSAA alone at only 3x GPU Scaling Factor was downright scary. We're only talking about doing MSAAx8 on a 768x576 framebuffer, and so an older not-so-terrible GPU should be able to handle this kind of load without having a total performance catastrophe.

But now let's get to the real issue I'm having...

@Jules-A, truth be told, your PR is blocking some changes that I want to make, and so to speed things up, how about this -- I'll just implement user-selectable MSAA sample sizes at the backend level myself, and also add the UI for this feature in the Cocoa port. You will be responsible for implementing the UI for the Windows port. How's that sound? I have some fairly extensive changes that I want to commit really soon, but I really need you to finish up with NDSSystem.h and frontend/windows/main.cpp first as to avoid some merging grief.

Jules-A:
> your PR is blocking some changes that I want to make
Oh sorry, I haven't got around to doing this yet, just WIP stuff. I was always prepared to deal with whatever changes were made in the meantime. If you are able to implement it at the backend that would be awesome as I was thinking about making some changes to show the currently used MSAA step in the UI for the Auto setting I was going to retain. I couldn't find out how to do this exactly without creating a global or modifying stuff I probably shouldn't. I'd be more than happy to just handle the Windows UI.

rogerman:
Okay, there you go. Commit 94ddf1a implements user-selectable MSAA sample sizes for the OpenGL backend.

So for frontends, simply set CommonSettings.GFX3D_Renderer_MultisampleSize to your desired MSAA sample value, and then the backend code will take care of the rest.

As for frontend elegance and niceness... that remains up to you...

Jules-A:
I did a [quick mock-up](https://user-images.githubusercontent.com/1760158/47323980-9ba95280-d690-11e8-971e-ddd38f1dba2d.JPG) of what it will likely look like now. Moved the "default" button (which oddly enough only defaults to software renderer up to more suitable location. 
I'm not sure if the new code changes anything but with the old code I was unable to get _deviceInfo.maxSamples from main.cpp. I'd like this so I can create combo-boxes only for the sizes the user supports. For now I need to sleep.

extherian:
> Ahh... much more reasonable to expectations -- thanks for doing this for me. Your report of 42% performance reduction with MSAA alone at only 3x GPU Scaling Factor was downright scary. We're only talking about doing MSAAx8 on a 768x576 framebuffer, and so an older not-so-terrible GPU should be able to handle this kind of load without having a total performance catastrophe.

There's no way an older GPU could handle something as intensive as 3x scaling. I use an i5 7600 and a 1050 ti, and for demanding games like Spirit Tracks my system can only barely handle 3x scaling with MSAA/24 bit color. On 4x with these settings I struggle to hit 75% of full speed in the opening of Spirit Tracks.

A couple of revisions ago, a menu appeared for selecting MSAA that has since disappeared, and with it set to 2x MSAA I could just about hit full speed on Spirit Tracks if I also set the color depth to 15 bit. Believe me, anything that can lessen DeSmuME's enormous performance deficit would be greatly appreciated by all DeSmuME users, as even users with modern desktop PCs struggle with some games.

rogerman:
> There's no way an older GPU could handle something as intensive as 3x scaling. I use an i5 7600 and a 1050 ti, and for demanding games like Spirit Tracks my system can only barely handle 3x scaling with MSAA/24 bit color. On 4x with these settings I struggle to hit 75% of full speed in the opening of Spirit Tracks.

@extherian: You are mistaken because your performance analysis is incorrect, as you are trying to relate MSAA performance or 3x scaling factor performance to overall graphics performance on a unique scene like the Zelda: Spirit Tracks opening scene. I have seen many low-end GPUs successfully run 3x scaling in many many games -- it's not as intensive a rendering mode as you think. The reason why YOU think it's too intensive is because you are cherry picking a literal worst-case scenario for the OpenGL renderer which has nothing to do with either MSAA or running at 3x scaling.

This scene is worst-case slow because this particular scene does a ton of texture thrashing, which forces a lot of performance-killing texture uploads to occur (texture uploading is required by OpenGL). This texture thrashing is further exacerbated if you have Texture Scaling Factor > 1 or Texture Deposterize enabled, as these options require CPU cycles to process the textures and also make the texture uploads even larger in size.

For best performance in this scene (and any other massively texture thrashing scenes), avoid the issues that are related to texture uploading or CPU post processing -- switch your 3D renderer to SoftRasterizer (does not require any texture uploads), change Texture Scaling Factor to 1 and turn off Texture Deposterize (avoids CPU post processing).

For Zelda: Spirit Tracks in particular, I highly recommend playing the game with SoftRasterizer because it avoids the performance pitfalls of texture uploading and polygon state validation (another performance pitfall on OpenGL).

Jules-A:
Okay so the interface is looking better: 
interface
But I can't seem to get the max device samples when the renderer isn't set to OpenGL using CurrentRenderer->GetDeviceInfo().maxSamples; (it works fine with OGL but null/SR return 0. 
I'm also not happy with the implementation of the checking for valid values entered from the ini. Currently it disables MSAA as my brain isn't working properly today.
Also not happy with interface creation but that's another situation of "if maxSamples was reliable...".

Jules-A:
>Create a WinOpenGL context for yourself and get maxSamples directly using glGetIntegerv(GL_MAX_SAMPLES, &maxSamples). You may want to take a look at frontend/windows/ogl.cpp to see if some context creation code already exists for your use. Getting maxSamples in this manner is fast, can be done at any time, and does not report irrelevant console stuff. (This is the method that the Cocoa port uses to get maxSamples on startup.)

If the frontends are all going to do this, shouldn't it be done in the backend?

rogerman:
> @rogerman I apologise for jumping to conclusions, I genuinely thought the 3D was just too demanding for my GPU to handle because the Spirit Tracks opening runs worse the higher I go above 3x. If it is the textures slowing me down, then I guess I must be bottlenecked by my CPU, not my GPU.

@extherian, DeSmuME in general is usually going to be bottlenecked by the CPU rather than the GPU. Also, the Spirit Tracks opening is a very unique rendering case that doesn't occur in the vast majority of games. Most games are well-behaved enough that they won't thrash textures. However, if a game does texture thrash, then on a hardware NDS, this doesn't matter because texture data is loaded directly from the game ROM, which is a very fast operation on an NDS. Trying to do the same thing in OpenGL involves transferring data across the PCI-E bus, which has significant latency and can be slow if the textures transferred are very large.

> It sounds like OpenGL is the culprit here as far as performance goes, if Softrasterizer doesn't have these performance issues. I wonder if Vulkan is any faster at texture uploading or polygon state validation. In Dolphin there are some edge cases that Vulkan is better at handling than OpenGL.

Don't be fooled by the Vulkan hype here! In the case of texture uploading, Vulkan is no faster than OpenGL because the performance issue is caused by repeatedly transferring large amounts of data across the PCI-E bus. But admittedly, polygon state validation would certainly be faster than OpenGL because Vulkan doesn't have to do so much of it. However, there are no plans to write a Vulkan 3D rendering backend at this time.

The only reason why SoftRasterizer can faster than OpenGL in certain rendering cases is because SoftRasterizer doesn't have to do any texture uploads (all data simply remains in main memory) and requires no state validation whatsoever. But when you encounter a game scene that does not texture thrash or require much polygon state validation (most games scenes), you will find that OpenGL is easily faster than SoftRasterizer, especially at the higher resolutions. If you want to see an example of OpenGL's performance superiority, try playing Kingdom Hearts: Re-coded and you will see it. Given your hardware (i5 7600 and GeForce 1050 Ti), you should be getting 60 FPS on a 5x GPU Scaling Factor with all the 3D graphics enhancements easily.

Jules-A:
Okay, everything seems to work correctly now, just not sure on if windows_opengl_init() fully destroys the context. There were a few things I was going to do differently if I had max samples now I've forgot. Either way I don't think I'll be doing anything more tonight.

Jules-A:
> if you try and change GPU Color Depth or Texture Scaling, the selection will be incorrect on one of the radio button groups.
Oh wow... wonder what happened there lol, might just have a quick look before I go to bed.
>Is there anything more that needs to be done?
Maybe just some code cleanups, it's looks a little hacky is all.

rogerman:
One more issue here... I noticed that the condition you use to prevent the MSAA pop-up menu from being generated more than once is to ensure that (maxSamples != 0).

However, it is possible that the user may be running an extremely old GPU that supports OpenGL but not Multisampled FBOs. Therefore, you should treat 0 as a legitimate return value for GL_MAX_SAMPLES and use a different condition for ensuring that the MSAA pop-up menu is generated only once.

Jules-A:
Okay so it's something for sure in resources.rc, just can't find what it is, need to look at it after sleep...
>However, it is possible that the user may be running an extremely old GPU that supports OpenGL but not Multisampled FBOs. Therefore, you should treat 0 as a legitimate return value for GL_MAX_SAMPLES and use a different condition for ensuring that the MSAA pop-up menu is generated only once.

Oh wow, okay then.

Jules-A:
Soo... I found the issue with the texture scaling and colour bits. For some odd reason it was the arrangement in resources.rc. I re-arranged everything so that it listed elements top to bottom, left to right for easier editing, turns out that was a bad idea... So I put everything back in the same illogical and not human readable positioning... 

Also added that extra check. I noticed a tiny bit of delay (nothing to worry about really) when opening the settings menu for the first time in non-OGL, I'm guessing that context creation isn't too cheap.

rogerman:
@Jules-A: Alright, I can confirm that the bug with the radio button groupings has been fixed.

> Also added that extra check. I noticed a tiny bit of delay (nothing to worry about really) when opening the settings menu for the first time in non-OGL, I'm guessing that context creation isn't too cheap.

The full context creation is NOT cheap because it has to set up every single struct and buffer in both main memory and VRAM for the 3D rendering emulation. That is why I recommended just creating a dummy context for the sake of getting GL_MAX_SAMPLES on startup because that is fast.

However, there's nothing wrong with your method of lazily getting GL_MAX_SAMPLES upon opening the 3D Settings window, just as long as you do it once and only once.

Which is why...

Rather than going through a lengthy diatribe of minor issues and stylistic quibbles, I'll just send you a main.cpp of what I'm looking for. Diff it and incorporate what you like. (It has extension .txt, but it is, in fact, the main.cpp file from src/frontend/windows/. GitHub doesn't like .cpp file extensions, so change it back to .cpp yourself.)

[main.txt](https://github.com/TASVideos/desmume/files/2512541/main.txt)

Jules-A:
@rogerman , I just copied it all though I'm a little concerned about placing didGetMaxSamples = true; AFTER running isTempContextCreated() since if it fails to get the context it will try every time the window is opened. It's probably not an issue, I just think it may be better placing it before since if it can't get an OGL context context it's still going to report maxSamples as 0;

rogerman:
If context creation fails, then maxSamples = 0 is intentional because then the MSAA pop-up menu will simply create a menu with only the "Disabled" option in it.

The placement of "didGetMaxSamples = true" AFTER isTempContextCreated is intentional because it gives the app a chance to retrieve GL_MAX_SAMPLES if it had previously failed to do so. Beforehand, even if context creation failed, then the app would NEVER attempt to retrieve GL_MAX_SAMPLES ever again, making it impossible for the app to recover.

Jules-A:
>it gives the app a chance to retrieve GL_MAX_SAMPLES if it had previously failed to do so.

Oh whoops. didn't think of that.

rogerman:
> Honestly the whole codebase needs a clean and defined style. I think I found 5 diff styles and I set the editor up to one part of the code only to find it completely different somewhere else and having to override the editor. I mostly follow what's surrounding but constantly changing just gets me confused.

Yeah, in general, you're right about that. However, I was referring to the following issues of your previous code (now that you've got me writing the diatribe):

1. Ensuring that maxSamples was retrieved once and only once (as long as no error occurred). Previous code would retrieve maxSamples every time the 3D Rendering Engine was an OpenGL one.

2. Eliminating the conditional statement (tempContextAttempted == false). As a stylistic rule, NEVER explicitly check a bool value for "true" or "false". For bool values, always use the ! operator instead.

3. Ensuring that at least the "Disabled" entry was always present in the MSAA pop-up menu, having no dependencies on any other variables. Previous code had dependencies on variables i and z, which could get changed in a future revision of the code, causing the pop-up menu to break.

Jules-A:
> Previous code would retrieve maxSamples every time the 3D Rendering Engine was an OpenGL one. 

I realise that now but that could be avoided by checking if maxSamples was 0 (at least for the cards that can use MSAA). I thought CurrentRenderer->GetDeviceInfo().maxSamples was a near 0 impact check to matter. It doesn't matter anyways since currently it's only a tiny slowdown slowdown that has a very low chance of happening only on subsequent setting loads. I'm not suggesting to change anything, I'm more questioning out of curiosity.

>As a stylistic rule, NEVER explicitly check a bool value for "true" or "false". For bool values, always use the ! operator instead.

Yeah that's where I get confused, I've seen it plenty of times in the Desmume code base so and rarely ever saw ! used for condition checks, at least not when dealing with multiple conditions. Normally I do use ! though...

>3.

Yeah I knew I was missing something, I even had a comment there to remind me lol.

Jules-A:
>Yes, there were exactly 2 instances of an "== true" statement in NDSSystem.cpp

[It's everywhere in main.cpp](https://user-images.githubusercontent.com/1760158/47468796-05139780-d830-11e8-9f38-12b58b8caf8d.JPG) the file I've mainly been in.

> In any case, can we merge this PR yet? 

Yeah.

rogerman:
> It's everywhere in main.cpp the file I've mainly been in.

Oh, ha ha... heh. I only checked the core code base, not the Windows frontend-specific code. Well... I guess this just chalks up another reason why the Windows frontend code needs a total rewrite.

As for merging this PR, I've gotta go someplace in a few minutes and I won't be back for several hours. I'll merge this PR when I get back, assuming that @zeromus doesn't beat me to merging it himself.

zeromus:
> The other instances of "== true" or "== false" statements are in external libraries, not the DeSmuME code base -- not our fault. But no matter what, it is bad form to do this.

`lhs==true` requires that the lhs be a bool. If you have a struct with a bunch of hetereogeneously typed members bearing names that don't give consistent indications of what types they are, using `lhs==true` consistently will keep you from accidentally evaluating a string as a bool. Look at it this way: it suppresses implicit evaluation of non-bool types as bools. That's a cool feature sometimes; not others. Don't be so hasty to label a sane idiom as bad form.

merge it whenever you want

Pik-a-boo:
AMD OpenGL driver support sucks so much on Windows.
Its almost like its non-existent at times

Jules-A:
@Pik-a-boo 
Yeah it's true that AMD's Windows OpenGL drivers are poor in comparison to Nvidia but there are so few games/apps that use OpenGL with it expected to decrease even further in favor of Vulkan adoption. While it would be nice to see AMD do a major improvement in that area, I really doubt that will happen, my only hope is that they fix bugs faster... 

While implementing a DirectX or Vulkan renderer in Desmume might greatly improve performance on AMD GPUs and bring it inline with it's Nvidia counterparts, that's a massive task that I doubt anyone will end up doing. There may also be newer OpenGL extensions that are able to bring performance closer to Nvidia's but it still requires the knowledge and time to implement.

zeromus:
it's not that massive a task. opengl is massive because it's a moving target, and rogerman wants to keep the old opengl versions working as well as possible. I don't know how much of a moving target vulkan is, but directx 11 isn't a moving target, and we might get directx 12 on windows 7 once they finish shaking the bugs out of it. I would expect a good-enough-to-be-debugged dx11 or dx12 renderer to be not even a week's worth of work for someone who knows what they're doing. Just base it on the most advanced opengl codepath.

PR #216, opened 2018-11-02
Win Port: Wifi cleanups following 9d05935

Jules-A:
This removes the the interface for setting wifi mode along with removing redundant code. Needs testing.
I merged the <0 and == 0 checks as they were pretty much the same message.


Jules-A:
Yeah, I knew that but it really isn't that helpful to the user. It needs a little more clarification but I'm not sure what else to say.

rogerman:
I'm hoping users interpret it like this:

"Error: Cannot find any devices." -- Means that an error occurred, and that winpcap attempted to find a network device but failed to do so.

"No devices found." -- Means that winpcap tried to find a network device but discovered that there were none on the system. This implies that winpcap worked perfectly and that no error occurred.

Maybe write "No devices found." to "No devices were found." to better clarify that winpcap legitimately found 0 network devices. I don't know how else to write this to be more meaningful.

One other thing... the datatype for "enableWin" should be a Windows-style BOOL that uses TRUE/FALSE rather than a C99 bool that uses true/false. Please use the proper datatypes when dealing with client APIs such as EnableWindow(), since you should assume that client APIs cannot be trusted to deal with datatypes that are different from their own.

Jules-A:
What about "Error: Failed while searching for a device." and "No devices found."?

PR #224, opened 2018-11-21
WIP WinFrontend UI for the new OpenGL compatabilty tradeoffs.

Jules-A:
Should be fully functional except for CheckValid3DSettingInt where I messed up so I left it commented out, honestly it's probably not needed anyways as there's already checks in place, it just doesn't update the ini until you save settings. UI is still a mess still, will have to work on that later.

Submitting as a WIP PR so people can test using using the automatic build. If you can help me understand what I'm doing wrong in CheckValid3DSettingInt it would be awesome but I'll probably figure it out after some sleep...


rogerman:
Just looked at the new UI. Everything seems to be functional, but it could look better.

The box that says "Trade Accuracy For Performance" is really an eyesore and doesn't really do anything functionally. It would be better to put the four checkboxes within a group box and label that group box "Rendering Accuracy Options."

And as an aside, I've noticed that all of these text descriptions beneath certain checkboxes are starting to clutter up the UI. In order to clean up the UI a bit, it would be better to start using tooltips for each control instead. This is what the Cocoa port does to keep the UI from getting out of control. As an example...

cocoaport3drenderingsettings

Of course, cleaning up the 3D Rendering Settings window isn't the purpose of this particular PR, so you don't have to do any cleanup if you don't want to. I'm just showing an example of what a nice-looking UI could look like.

Jules-A:
Yeah tooltips would be nice but I was leaning more towards Dolphin's solution as it would allow more detail (tooltips can also be annoying sometimes). That being said I don't really have much time to work on that yet so save for another PR. About that weird inverted space, that is Microsoft's definition of a "border"... I replaced it with group box but inside OpenGL box since they obviously are only for OpenGL but I'm thinking maybe just having normal text there instead would be better.

rogerman:
Ah, nice. The group box looks much better now. But all the UI elements seem to run together and the user loses the visual groupings of each option. Try tightening up the spacing like this:

- For each option, nudge each description text closer up to their respective checkboxes by 1px.
- For the "Enable Depth Equals Test Tolerance" and "Enable Depth L-Equal Polygon Facing" options, nudge their respective checkboxes and description texts down by 2px.

Also, if the "TT" in the symbol name "ID_LABEL_DEPTH_EQUALS_TT" stands for "Test Tolerance", then shouldn't the symbol name "ID_LABEL_DEPTH_LEQUALS_TT" actually be named "ID_LABEL_DEPTH_LEQUALS_PF", where the "PF" stands for "Polygon Facing"?

Jules-A:
Ughh... copy and paste fail... I made the names a bit different this time and fixed up spacing. I finally managed to get ini sanity checking working without it being a bloated mess as well.

rogerman:
Just checked the latest changes. Everything looks good now.

Go ahead and make another commit to remove the "bool valid = false;" statement. I will merge the PR after you do that.

Jules-A:
Oh, I also changed the combobox generation. Basically since it gets initialized with defaults in Commonsettings, checked in cmdline and now checked in ini, it's not possible for the value to be incorrect.

PR #233, opened 2018-12-02
[Windows] Hide the console window in the current session when disabling in menu.

Jules-A:
Currently under Windows 1803 and 1809 (maybe other windows versions as well), when disabling console, it will only detach from the Desmume process and continue to stay open and visible. Since I don't really know much about the console and how to fix it, we should just hide the window for now.


zeromus:
ok whatever you say

PR #239, opened 2018-12-18
[Windows]Use 64bit linker as backup instead of failing to find

Jules-A:
Recently appveyor and my own builds have been failing fairly often with "out of heap-space" type errors. It tries to find the 64bit linker but it's not configured to find it within the project.

I've only tested this with VS2017 and the output is created from within VS property manager so I'm not sure if it works on older compilers but I guess by submitting a PR it tests appveyor's vs2015.

### DIFF
--- a/desmume/src/frontend/windows/DeSmuME.vcxproj
+++ b/desmume/src/frontend/windows/DeSmuME.vcxproj
@@ -58,6 +58,10 @@
     <Import Project="desmume.props" />
   </ImportGroup>
   <PropertyGroup Label="UserMacros" />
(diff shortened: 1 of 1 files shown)


zeromus:
I thought about it and decided I just wanted to use the 64bit tools instead. If it can't build reliably on 32bit tools, let's just be clear that it requires the 64bit tools from the beginning (and pushed a commit to do this)

Jules-A:
Yeah that makes sense, Microsoft say the only reason they use the 32bit by default is because the 32bit is faster and that most don't need 64bit.

PR #251, opened 2019-01-28
Fix broken code.

axzxc1236:
I'm pretty sure what I replaced is not C, so I ported it to C, and it compiled.


zeromus:
It's included by `#include <streams/file_stream.h>` in that source file.
You didn't fix the problem correctly.

axzxc1236:
Okay... I don't have a clue on how to fix it correctly.
Just want to let you know, I got undefined reference to "filestream_open", "filestream_gets" and "filestream_close" when I tried to compile the code.

PR #273, opened 2019-05-04
Add Lua function gui.setlayermask

awforsythe:
This change adds a Lua function called `gui.setlayermask`, which allows Lua scripts to procedurally modify which render layers are visible. It takes two arguments: one integer bitfield representing the visibility states for the main GPU, and the other representing the sub GPU. For example: `0b11111` shows all layers, `0b00000` hides all layers, `0b00001` shows only layer 0 (`BG0`), `0b10000` shows only layer 5 (`OBJ`), etc.

Since display layer state is coupled to the frontend, and since frontends are entirely separate between platforms _(i.e., on Windows, toggling display layer state requires updating the GUI state as well)_, this function is only supported and tested on the Windows build. On MacOS and Linux, `gui.setlayermask` will simply return 0 without doing anything.

I added this function because I wanted to be able to write Lua scripts that can capture screenshots with specific layers isolated. That project (with example usage in `nds/lua/lib/screenshot_task.lua`) can be seen here: https://github.com/awforsythe/rando-pokedex

I'd like to get this change merged into DeSmuME so I can publish that project without also having to support a custom branch of the emulator. I'd love to get the same functionality working on the MacOS and Linux frontends so that there aren't disparities in Lua functionality between platforms, but I don't have the resources to build and test on those platforms at the moment. If there are better ways of accomplishing the same thing, or better / more conformant approaches you think I should take, I'm open to feedback.

Thanks!

### DIFF
--- a/desmume/src/frontend/windows/display.cpp
+++ b/desmume/src/frontend/windows/display.cpp
@@ -968,3 +968,28 @@ void TwiddleLayer(UINT ctlid, int core, int layer)
 	gpu->SetLayerEnableState(layer, newLayerState);
 	MainWindow->checkMenu(ctlid, newLayerState);
 }
+
+void SetLayerMasks(int mainEngineMask, int subEngineMask)
+{
+	static const size_t numCores = sizeof(CommonSettings.dispLayers) / sizeof(CommonSettings.dispLayers[0]);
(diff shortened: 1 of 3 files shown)


zeromus:
dont worry about the other platforms, this is low impact enough that whatever gets the job done is adequate

PR #276, opened 2019-05-07
typo related updates

fpscan:
Since full text unable to read when it was `35`, changed to `71`, so its readable right now.
Typo corrected
Added Windows Border to sound setting to make it more important
“Wi-fi” [is used to certify the interoperability of wireless computer networking devices](https://www.merriam-webster.com/dictionary/Wi-Fi). So `Wifi `changed to `Wi-fi`

### DIFF
--- a/desmume/src/frontend/windows/resources.rc
+++ b/desmume/src/frontend/windows/resources.rc
@@ -1020,7 +1020,7 @@ CAPTION "Sound Settings"
 FONT 8, "MS Sans Serif", 0, 0, 0x1
 BEGIN
     GROUPBOX        "Synchronization Mode",IDC_STATIC,13,8,173,80
-    CONTROL         "Dual SPU Synch/Asynch (traditional)",IDC_SYNCHMODE_DUAL,
+    CONTROL         "Dual SPU Synch/Async (traditional)",IDC_SYNCHMODE_DUAL,
                     "Button",BS_AUTORADIOBUTTON | WS_GROUP,22,18,132,10
     CONTROL         "Synchronous (sometimes needed for streams)",IDC_SYNCHMODE_SYNCH,
                     "Button",BS_AUTORADIOBUTTON,21,28,158,10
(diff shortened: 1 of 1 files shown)


zeromus:
Revert the wifi change. Unwanted. Not going to debate. But for your own reference, grep the source code for wifi and wi-fi and compare the results and think about that.

17WS_BORDER looks like a mistake?

fpscan:
> Revert the wifi change. Unwanted. Not going to debate. But for your own reference, grep the source code for wifi and wi-fi and compare the results and think about that.
> 
> 17WS_BORDER looks like a mistake?

No worries at all, I added `WS_BORDER ` to make it more eye-catching since it is important but not really necessary, if you like to I could delete that as well?

zeromus:
Uhhh I just mean, 17WS_BORDER looks like you meant to replace 17 with WS_BORDER but combined them instead.
And anyway, the border for that dialog shouldn't be any more or less eye-catching than any of the other similar config dialogs

fpscan:
> Uhhh I just mean, 17WS_BORDER looks like you meant to replace 17 with WS_BORDER but combined them instead.
> And anyway, the border for that dialog shouldn't be any more or less eye-catching than any of the other similar config dialogs

Yeah, so Border and Wifi changes removed

desmume/src/frontend/windows/resources.rc
```diff
@@ -1430,7 +1430,7 @@ IDD_SLOT1_R4 DIALOGEX 7, 64, 302, 89
 STYLE DS_SETFONT | DS_FIXEDSYS | WS_CHILD | WS_VISIBLE | WS_SYSMENU
 FONT 8, "MS Shell Dlg", 400, 0, 0x1
 BEGIN
-    CONTROL         "Directory to scan:",IDC_R4_FOLDER,"Button",BS_AUTORADIOBUTTON | WS_TABSTOP,9,22,35,10
+    CONTROL         "Directory to scan:",IDC_R4_FOLDER,"Button",BS_AUTORADIOBUTTON | WS_TABSTOP,9,22,71,10
```
fpscan: it wasn't readable with `35`, so changing to `71` make it visible and readable

PR #281, opened 2019-06-01
Add gui.savescreenshotas

martmists:


zeromus:
Forgot to commit hotkey.*

martmists:
Hotkey.h is already included though, I'm using the DoScreenshot function made by you guys defined [right here](https://github.com/TASVideos/desmume/blob/master/desmume/src/frontend/windows/hotkey.cpp#L162)

zeromus:
you _edited_ hotkey.*

Your commit is also probably bugged because DoScreenshot doesnt exist on non-windows ports; this would break the non-windows ports

martmists:
I did not touch hotkey.* though, and I'll see how I can fix it being windows-only

zeromus:
You did, because DoScreenshot() is static in hotkey.cpp. You removed the static and added the declaration to hotkey.h, then added #include hotkey.h so that the DoScreenshot declaration was available to the lua-engine.

martmists:
I'll add the declaration to hotkey.h then

zeromus:
No, if you haven't already edited hotkey.h, then figure out why you made a PR without BUILDING it first. Since the code you PR'd can't possibly build. Also figure out why you added #include "hotkey.h" if you didnt MODIFY hotkey.h, or if it wasn't required to make it build (you wouldn't know what was required to make it build, because it never built for you)

PR #294, opened 2019-09-06
GTK: add Auto pause in background option

indivisible:
Add option to automatically pause emulation when the main window loses focus for the GTK frontend.

The option is disabled by default, and can be enabled in `Config > Auto pause in background`. It is persisted in the config as `[Window] > AutoPause`.

Receiving focus will only resume the emulation if it was stopped automatically, so manually pausing will not be undone unintentionally.


zeromus:
Sorry, this seems like it's probably obsolete now that we have someone doing a lot of GTK work. If you want to repeat the patch I'll merge it this time since there's a hope for it getting maintained.

PR #323, opened 2020-04-23
Fix glade UI file path in local builds

lucianposton:
Update `GLADEUI_UNINSTALLED_DIR` to work with the current directory layout.

Also adds a failure message when the glade ui file is not found.


lucianposton:
The CI failure appears unrelated.

Issue #202, opened 2018-10-07
LINUX - Segfault when trying to switch to OpenGL


blacksparrow123:
I'm on manjaro.
Desmume installed via AUR.

zeromus:
let's play a guessing game. I'm thinking of the thing everyone asks for when all you say is "I got a segfault". Can you guess what I'm thinking?
otherwise--sounds like a question for the maintainer of the desmume package, which isn't us.

blacksparrow123:
Yeah. Not very informative.
Does desmume have some sort of logs or something?

I'll try to contact them if they are still around.

zeromus:
oh, supposedly this was fixed in may. good enough for me

PR #327, opened 2020-05-03
Fix unhandled null causing crash when using osmesa with nvidia driver

### LINKED ISSUE #126: [Linux] Crash on Graphics OpenGL Selection
Opening the GUI, selecting OpenGL in Graphics Settings and then hitting OK results in the interface crashing. Here's the console output from when it happens. It then seems to crash immediately on the next startup too, but then works the next time - that appears to be because it's defaulted back to Software at this point.

```
CPU mode: Interpreter
Microphone successfully inited.
DeSmuME 0.9.12 git#0 x64-JIT SSE4.2+AVX2

Using 8 threads for video filter.
SoftRast Initialized with cores=8
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_M_construct null not valid
Aborted (core dumped)
```

This is built from GIT as of about 20 minutes ago. This is on a machine using the official Nvidia drivers with a GTX1060.

davg-qqq:
Should fix #126 and #202 
I'm worried that this might hide errors because it silently ignores them if it can't read an extension name. Maybe it's better to crash or log that the value was null?

### DIFF
--- a/desmume/src/OGLRender_3_2.cpp
+++ b/desmume/src/OGLRender_3_2.cpp
@@ -2033,8 +2033,11 @@ void OpenGLRenderer_3_2::GetExtensionSet(std::set<std::string> *oglExtensionSet)
 	glGetIntegerv(GL_NUM_EXTENSIONS, &extensionCount);
 	for (size_t i = 0; i < extensionCount; i++)
 	{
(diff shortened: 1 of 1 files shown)


zeromus:
Youre the one making edits to the linux port. You decide.

PR #330, opened 2020-05-08
Linux GTK Port: Fix config for new MSAA options

davg-qqq:
a289055e removed the code that updated the old "multisampling enabled"
config. This fix adds a new config value for the multisampling size and
updates the old config in sync. When loading the config, the new
multisampling size value is prioritized, but if it is 0 (possibly not set)
then it falls back to the previous logic using the old boolean config
value.


zeromus:
you now have commit access. anything that just touches the unix ports, push yourself. make a PR or get on IRC if you want to touch the emulation core

PR #347, opened 2020-07-20
gtk: add import/export backup memory menu items

prg318:
This branch adds the ability to import or export backup memory using the gtk frontend.


zeromus:
Sorry... we have someone doing major gtk commits now. If you repeat this work, you can push it directly rather than waiting on me to notice a PR

PR #350, opened 2020-07-31
DeSmuME as a shared object / dynamic library

Parakoopa:
This Pull Request adds a new "interface" frontend which compiles into a C ABI compatible DLL/SO, to be used by other applications. It compiles under macOS and Linux using the GNU build tools and for Windows I created a "fork" of the Visual Studio project.

It mainly revolves around `desmume/src/frontend/interface/interface.h` which should be fairly self-explanatory.

OpenGL is not supported under Windows yet (since I couldn't really figure out how to make it work). Some other features like OSD support and Lua are also not included, since they aren't really needed. The interface provides functions that would normally be usable by the Lua bindings, like memory hooks.

I put it on "Draft", because I definitely **still want to add a documentation** and clean up some things (some of the things I added for Windows might not actually be needed anymore, I need to check this when I have access to a Windows machine again). 
I create this more so that I don't forget. But feedback is already welcome.

I am not a C++ developer, this is literally the first thing I've ever done with C++, so if some things are bad, I apologize in advance.

This is used by py-desmume, Python bindings for DeSmuME: https://py-desmume.readthedocs.io/en/latest/
Which are currently used by a ROM hacking tool I'm working on: https://www.youtube.com/watch?v=_yD9OZobSoE


zeromus:
All I really care about is if it mucks anything else up, and it doesn't look like it would.

zeromus:
how's it going with your docs

Parakoopa:
I'll hopefully have it done by the end of the week.

evelikov:
Greeting everyone.

Have you considered using [meson](https://mesonbuild.com/) instead of autoconf + MSVC?

It works on Windows and Linux generating Make/Ninja/MSVC build files, while having more python-like syntax. It can download specific version of respective projects (SDL, 7z) so you no longer need to import headers/libraries into git. Plus using SDL2 is trivial - SDL1.2 is on death bed.

The existing posix backends [use it](https://github.com/TASVideos/desmume/blob/master/desmume/src/frontend/posix/meson.build) plus adding a CI job is [pretty trivial](https://github.com/TASVideos/desmume/blob/master/.github/workflows/build.yml) 

Complete outsider, so take my idea with a healthy pinch of salt.

P.S. I'm not affiliated in any way with meson

zeromus:
linux frontends already use meson. Windows frontend won't be changed.

evelikov:
> Windows frontend won't be changed.

I'm not saying it should. I'm suggesting that @Parakoopa may have easier time maintaining a meson build for the newly introduced "interface" frontend.

Parakoopa:
It might be something worth looking into as a next step, but for now I'm just happy it works haha. Both the autoconf and Visual Studio project are pretty easy to get running.

Parakoopa:
Added documentation and rebased.

Parakoopa:
I just noticed there was a migration to SDL2. I will have to make some adjustments to the posix build and probably change some things in the Windows build as well.

Parakoopa:
@linkmauve I see you replaced the posix build process with meson and also migrated to SDL2. Coul you give me a hand in making this compatible with your changes and possibly migrating to meson as well?

Parakoopa:
I switched to Meson for the posix build and updated both to SDL2. Should be ready now.

Parakoopa:
@zeromus
I cannot build the interface or the normal Windows version on Github CI runners. The link time code generation optimization takes too much memory.
I have also observed this locally, something very strange is going on there. It seems mostly related to the GPU module.
Without these optimizations the emulator runs at less than 1FPS under Windows.

Because of this I might look into building the Windows version of the DLL also with gcc. I'm really not sure what is happening there.

zeromus:
It's expected to take a long time. If your system or patience can't handle it, use the Release Fastbuild configuration in the vcxproj.

Parakoopa:
Right, thank you, I didn't consider that option.

Parakoopa:
I added Github actions to build the DLL/SO on Linux, Mac and Windows (x86/x64). Now this is truly ready to be merged :)

zeromus:
I spot checked it and it seemed to have minimal impact on anything else so I have no complaints. I'll take responsibility if anything got broken and you're awol. Thanks for your contribution

zeromus:
Can you tell me why an SDL is checked in here? doesn't seem to me that a desmume built as a dll with an API should have anything to do with SDL

Parakoopa:
Mostly for the input / joystick processing that was already available for the posix interface and the interface as an option to draw and process input using an SDL management window, behaving much like the posix command line interface.
This is mostly for people wanting to use the interface for simple prototyping or reverse engineering work without actually having to worry about I/O too much.

zeromus:
If we are going to have an officially checked in "interface" style project then it needs to be "headless" since that's the holy grail for this. No dependencies on I/O. Instead what we should have is an example client that incorporates the I/O. I don't think I/O should be a "feature" of the interface. You may as well expose an API on the full executable then (as has been done with lua, or could be done by sockets or similar).

Parakoopa:
Maybe as a compromise, SDL related features and the dependency on SDL could be disabled via a compile flag?

zeromus:
I feel strongly that an interface library should have a REFERENCE IMPLEMENTATION CLIENT. The two go together like peanut butter and jelly. You built the reference client implementation into the library. That's really weird. An articulatable cost of that is you've polluted the `desmume_*` api namespace with stuff not really related to desmume but for showing windows and stuff..... it's... so... weird....

Parakoopa:
I don't have any prior experience building something like this and said so in the original Pull Request description. I originally built this mostly to be useful for my purposes.

zeromus:
OK. I'll think about it. Stay tuned.

PR #367, opened 2020-08-31
Give thread names to ease debugging

linkmauve:
This makes it much more easy to figure out why a given thread is consuming CPU time for instance.

This is still missing sound threads, but those depend on SDL apparently.

### DIFF
--- a/desmume/src/GPU.cpp
+++ b/desmume/src/GPU.cpp
@@ -1414,7 +1414,7 @@ GPUEngineBase::GPUEngineBase()
 	if (CommonSettings.num_cores > 1)
 	{
 		_asyncClearTask = new Task;
-		_asyncClearTask->start(false);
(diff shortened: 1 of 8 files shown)


zeromus:
update rthreads from libretro-common before revising it

linkmauve:
It is now merged.  How do you usually deal with synchronising your vendor’d dependencies?

zeromus:
paste from vendor and check diffs to make sure there are not important local changes clobbered

zeromus:
what's the deal with this? have are we synchronized with libretro-common yet?

zeromus:
let's merge this and if it breaks the build, someone will have to update rthreads

MaddTheSane:
…And breaks macOS: Mac's `pthread_setname_np` takes only one argument: the new thread name. And it can only be called from the thread you want to rename.

A quick and dirty fix is in #422.

Issue #209, opened 2018-10-26
Wishlist for linux port obvious features


zeromus:
Please dump your obvious feature requests in here. We don't need 1000 individual tickets for what amounts to "support everything nice that's supported by other platforms"

zeromus:
#209 - add hotkey configuration

Jules-A:
Linux port also got left out on the MSAA settings, more information: https://github.com/TASVideos/desmume/commit/2e7099052d714d366bcdda51de61e52b5387f9d6

v1993:
@zeromus Please, fix link for hotkeys bug (#209 → #208).

dnmodder:
Allow to change the console language from the interface

dmgolembiowski:
Literally just documentation.

feuerpanda:
the LUA scripting part. Learning with Pokemon. Typing Adventure to be playable needs lua to have this script https://github.com/AnimeCommander/Learn.With.Pokemon-Typing-Adventure.lua

trymeouteh:
Please make the linux version have more features and settings like the windows version has such as...

Screen gap thickness and color
select firmware files and bios images
firmware settings
hotkeys
Boot from firmware option (to boot DS menu with game loaded instead of booting into the game)
Separate folders for the saves and savestates (Windows has the Battery and States folder)

dylmeadows:
I'm looking into trying to add PrescaleHD to the linux version. Is there anyone that can help steer me in the right direction?

zeromus:
Don't discuss this here, it's a thread hijack, and if you need any steering at all starting from the beginning, you're going to need your hand held the whole way.

sl1pkn07:
switch to gtk3 or Qt

trymeouteh:
> switch GUI to gtk3 or Qt

I would like gtk3 but thats just a personal preference

lowendgamer:
> Please make the linux version have more features and settings like the windows version has such as...
> 
> * Screen gap thickness and color
> * select firmware files and bios images (ARM7, ARM9, firmware.bin)
> * firmware settings (Change username, color, birthday)
> * hotkeys (hotkey menu editor)
> * Boot from firmware option (to boot DS menu with game loaded instead of booting into the game)
> * Separate folders for the saves and savestates (Windows has the Battery and States folder)
> * 7zip file support for games. (Open 7z file games)
[...]

PR #368, opened 2020-09-04
Port DeSmuME to GTK+ 3.24

linkmauve:
My end goal is GTK 4 support, but this seems like a good first step.

I’ve also added a meson build system, as it’s much more developer-friendly than autotools, and is getting widely adopted.

Please test the various features, amongst known broken is the recent file list (but you can use the recent file opener in the GTK+ open dialog).

I have only tested on Linux, but I see no reason it wouldn’t work elsewhere too, especially since meson supports Visual Studio.

Amongst my future plans for this GUI are:
- Port it to GTK 4, as said above.
- Reformat the code, probably using `clang-format`.
- Split the menu and `GtkBuilder` strings out in other files, possibly using `GResource` to keep them inside of the final executable.
- Replace the GLX context with a `GtkGLArea`, so that it works on Wayland without requiring Xwayland.
- Upgrade SDL to SDL2, for the same reason.

Fixes part of #209.


zeromus:
Nobody is going to test a fork, you will have to push it to master for anyone to test it. Then the test will be in the form of people reporting regressions as github bugs. So fix the bugs you know about and then I'll merge it.

linkmauve:
Here, quick savestate access from the menu, as well as recent files are now working properly.

I believe this could get merged now.

PR #385, opened 2020-10-02
GTK: Fix gdb stub support

### LINKED ISSUE #381: GDB Stub not working
While the meson option `-Dgdb-stub=true` causes an extra file to be built, it never defines the GDB_STUB variable, causing support for it to not completely be enabled.

EDIT: A `--enable-gdb-stub` build with autotools doesn't work either, due to variable scoping. Even after fixing this, the `--arm9stub` option isn't recognized in the command-line, despite being present in the `--help`. What happened to break this so badly? This was all working not _that_ long ago...

linkmauve:
I never tried to build this codepath, this is now done.

Fixes #381.

### DIFF
--- a/desmume/src/frontend/posix/gtk/main.cpp
+++ b/desmume/src/frontend/posix/gtk/main.cpp
@@ -104,6 +104,11 @@ VideoFilter* video;
 
 desmume::config::Config config;
 
+#ifdef GDB_STUB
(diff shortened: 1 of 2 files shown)


zeromus:
you know you can push this stuff yourself now

linkmauve:
Oh right, I was also fixing stuff in other projects and it’s a workflow I’m more used to. Sorry.

PR #394, opened 2020-11-07
GTK: Update autotools for new icon and desktop names

h3xx:
Meson was updated, but not the autotools build.


linkmauve:
I’m fine with that, but I was thinking about removing the autotools build system, we don’t need two of them.

I was also thinking about removing the gtk-glade UI, which is clearly abandoned.

h3xx:
To answer your question, I had a packaging script that used autotools. But I updated it today to use meson (hence #395), and so no, I don't personally need the autotools build system.

linkmauve:
Thanks, it’s now removed. :)

PR #395, opened 2020-11-07
GTK: Use standard meson to install man pages

h3xx:
Use standard function `install_man` which automatically determines the mandir and `man[0-9]` instead of guessing. Now it respects the `--mandir` option.

By default it's set to `--mandir=share/man`, which is where it was installing.


linkmauve:
Thanks!

PR #400, opened 2020-11-22
GTK: fix GL context creation with prime render offload (nvidia)

### LINKED ISSUE #399: OpenGL init failed on nvidia optimus card
This started to happen after update to SDL from GLX, OpenGL init successful on integrated graphics but fails on the second card (nvidia). Other SDL apps work fine.

thesourcehim:
Generally it's not recommended to force visual buffer format but instead let SDL decide which one is better, because not all formats are available on cards without screens.
Let's just add some diagnostics messages after GL context creation.
Fixes #399

### DIFF
--- a/desmume/src/frontend/posix/gtk/sdl_3Demu.cpp
+++ b/desmume/src/frontend/posix/gtk/sdl_3Demu.cpp
@@ -49,30 +49,34 @@ bool deinit_sdl_3Demu(void)
 
 bool init_sdl_3Demu(void) 
 {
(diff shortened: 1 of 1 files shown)


thesourcehim:
Tested more and it seems that auto detection creates color buffer without alpha channel. However, it also seems, that those attributes should be set before SDL_CreateWindow to work correctly with optimus. I don't know why it matters, but now it works with optimus with all attributes set.

linkmauve:
Hi, I pushed these commits with minor changes, please report if everything works on Nvidia now.

PR #403, opened 2020-11-26
Change DS window background to black

anlsh:
See https://github.com/TASVideos/desmume/issues/40


zeromus:
what was the color before? 
please redo on latest commit and ill merge

PR #422, opened 2021-01-25
Update rthreads.c

MaddTheSane:
Make it build again on macOS.

### DIFF
--- a/desmume/src/libretro-common/rthreads/rthreads.c
+++ b/desmume/src/libretro-common/rthreads/rthreads.c
@@ -314,8 +314,8 @@ void sthread_setname(sthread_t *thread, const char *name)
 {
    if (!thread)
       return;
-   // TODO: implement that for Windows too.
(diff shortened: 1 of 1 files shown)


zeromus:
update rthreads from libretro instead of making our own edits to it. the guy who did this sent it to libretro and it was merged, so theyve surely fixed whatever problems are in it since then

MaddTheSane:
Libretro from https://github.com/libretro/libretro-common? It looks like the thread naming function was removed in https://github.com/libretro/libretro-common/pull/168.
…although there is a new pull request (https://github.com/libretro/libretro-common/pull/169) that would do the same thing better than this patch (limits it to only Linux; NetBSD and OpenBSD have [different calling conventions](https://stackoverflow.com/a/7989973/1975001) than Linux).

zeromus:
ok thanks for doing that legwork

PR #436, opened 2021-03-13
GTK: fix freeze when editing joystick controls

thesourcehim:
Background events for joysticks need to be explicitly enabled, otherwise they only work when SDL window has focus apparently.


linkmauve (approved):
LGTM, thanks!

PR #439, opened 2021-03-20
Update arm9 and arm7gdb parameters.

PlatinumMaster:
The README used the older version of the syntax, (e.g `--arm9gdb=PORTNUM`).


zeromus:
I'm not gonna merge this without ALL the options being fixed in the docs. You don't have to confirm them all. You can simply compare to the --help output to see that NO equals signs are used in 2021

PlatinumMaster:
That's fine with me. I will fix these soon.

PR #466, opened 2021-09-14
Change background color on Linux

anlsh:
This PR updates #403 for the current desmume-gtk (Linux) code.

Up until now, the Desmume looked like the picture below on my machine (and perhaps Linux machines in general)
before-update

The light backgrounds were an eyesore. This commit changes the background color to black and the status bar text to a light gray so that it is readable on the black background.
after-update.

It's obviously not perfect: the toolbar is still light-themed (I hope to fix this at some point by defaulting to a dark GTK theme), and the buttons still become a whiteish color on hover (a theming issue). However the current solution is good enough in fullscreen, where the clashing elements are hidden. I was hoping to get this merged for now

### DIFF
--- a/desmume/src/frontend/posix/gtk/main.cpp
+++ b/desmume/src/frontend/posix/gtk/main.cpp
@@ -507,6 +507,7 @@ static GtkWidget *pWindow;
 static GtkWidget *pToolBar;
 static GtkWidget *pStatusBar;
 static GtkWidget *pDrawingArea;
+static GtkWidget *pContentBox;
 
 struct nds_screen_t {
     guint gap_size;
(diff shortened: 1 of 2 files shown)


zeromus:
That's eccentric. Normally status bars and toolbars are themed as the window chrome and not devoured by the interior blackness.

anlsh:
desmume-gtk does not seem to respect themes: I have tried launching it with the appropriate environment variable set, to no avail. [Others](https://forums.lutris.net/t/white-bar-letterbox-in-desmume/2375/4) seem to have encountered this issue as well.

I wouldn't call this change "eccentric" either: in fact it seems to be a quite popular ask in general, with a search for "Desmume dark theme gnome" showing many results of people asking for black backgrounds ([1](https://forums.launchbox-app.com/topic/52978-desmume-fullscreen-white-edges-to-black/), [2](https://forums.lutris.net/t/white-bar-letterbox-in-desmume/2375), [3](https://www.reddit.com/r/Roms/comments/o4v9cw/desmume_emulator_dark_mode/), [4](https://www.reddit.com/r/PokemonROMhacks/comments/jwx8uw/get_rid_of_annoying_white_background_for_desmume/), [5](http://forums.desmume.org/viewtopic.php?id=11801)). Windows users often ask this exact question as well, and are told that they can force a black background by setting their renderer to OpenGL: a solution which does not work on Linux, at least on my machine.

[You yourself](http://forums.desmume.org/viewtopic.php?pid=23485) have explicitly said that "they [users] want black [backgrounds]," and were open to my outdated PR (#403) addressing the same issue. Again, while my PR is clearly imperfect it is still an improvement on the current status quo for Linux users: and as it will only affect the GTK build, I think it merits a merge

zeromus:
I appreciate the great lengths you went to to zing me, I really do. However 
1. here I was talking about the statusbar and the toolbar, not the backdrop in the main window client area. The clue was that I said "status bars and toolbars".
2. "I myself once said" something about the backdrop in the window client area.

anlsh:
I've updated the PR so that the second layer of the toolbar isn't given a black background: you can see what it looks like below
eccentric

I actually think the status bar looks better like it does here: the text is clearly legible, and the black background means that it doesn't clash with the client area. If you consider still that to be an issue, I can restore the light background

zeromus:
OK I still think it's eccentric but now that you properly understand the situation, you can do it any way you want.

PR #472, opened 2021-09-21
Disassembler: Fix LDR/STR immediate offset

MathyFurret:


zeromus:
Uhmmm I dont know you and theres not a lot on github to stalk you through. can you just post a reference from an arm doc or gbatek or another emulator sources? I'm not saying what we've got is right, but a lot of smart people have used it and not found this problem.

MathyFurret:
LDR/STR(B) uses a 12 bit offset https://problemkaputt.de/gbatek.htm#armopcodesmemorysingledatatransferldrstrpld

The actual emulator defines it correctly https://github.com/TASVideos/desmume/blob/master/desmume/src/arm_instructions.cpp#L249

PR #480, opened 2021-10-29
restore autotools build sys and GTK+2 frontend

rofl0r:
- readded last working GTK+2 frontend, now in directory gtk2 in order not to clash with gtk+3 one
- restored and fixed autotools (only for gtk2 and cli frontend, gtk+3 frontend in "gtk" dir is still meson-oly, and cli can be built with both)
- fix build with musl libc
- added 2 new functions to gtk+2 port (memory dump and import backup)


rofl0r:
a growing number of people decided to stop playing the forced planned obscolescence dance.
https://sabotage-linux.neocities.org/blog/8/

PR #481, opened 2021-10-31
Update menu on frameskip value change

thesourcehim:
Make Modify_FrameSkip behave like ToggleAutoFrameSkip and ToggleFpsLimiter, i.e update menu state and print new value via driver->AddLine


thesourcehim:
Some how an old commit is displayed here too despite being merged earlier.

PR #485, opened 2021-11-04
interface: Fix macOS build

tech-ticks:
The build doesn't load on macOS due to missing symbols in the outdated the system `libstdc++`. `clang` uses the non-deprecated `libc++` instead, which fixes issues when loading the library. Also properly references the OpenGL dependency.


desmume/src/frontend/posix/shared/mic_openal.cpp
```diff
@@ -1,25 +1,30 @@
...
-	(at your option) any later version.
-
-	This file is distributed in the hope that it will be useful,
-	but WITHOUT ANY WARRANTY; without even the implied warranty of
-	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-	GNU General Public License for more details.
-
-	You should have received a copy of the GNU General Public License
-	along with the this software.  If not, see <http://www.gnu.org/licenses/>.
+	Copyright (C) 2009-2010 DeSmuME team
```
rofl0r, line 2: it appears your editor somehow produced gratuitous whitespace changes (if you use `git add --patch filename` you could exclude the hunk)

tech-ticks: Turns out this file is using Windows line endings in this comment and Unix line endings everywhere else. Would it be alright to commit it like this to fix the inconsistency? If not, I can try to configure Git to commit to commit the line endings without changing them, but a simple `git add --patch` doesn't help here.

rofl0r: > a simple git add --patch doesn't help here.

yeah, you'd need to unstage the commit beforehand and commit again.

> Turns out this file is using Windows line endings in this comment and Unix line endings everywhere else.

yeah' there's a few such inconsistencies in the codebase, my editor also "fixed" a couple unrelated lines when editing some files, though i noticed before commiting by looking at the changes with `gitk`, so i used git add --patch to exclude those hunks. but yeah, i'm personally ok with having the whitespace fixed in this case, it would be nicer if it was a separate commit but let's not be too bureaucratic. i don't think @zeromus cares either.

tech-ticks: Alright, I've restored the original file endings. FYI in case anyone else has this issue in the future, unstaging and commiting again didn't work until I've disabled auto CRLF with `git config --global core.autocrlf false`.

desmume/src/frontend/posix/shared/mic_openal.cpp
```diff
@@ -18,8 +18,13 @@
 
 #ifndef WIN32
 
-#include <AL/al.h>
-#include <AL/alc.h>
+#ifdef __APPLE__
+  #include <OpenAL/al.h>
+  #include <OpenAL/alc.h>
```
rofl0r, line 23: out of curiosity... why is this different on apple, and why did it work in the past ? is it system libs vs homebrew ?

tech-ticks: I don't think anyone has ever trie to build interface under macOS with `clang` before. Maybe it's different when compiling with `gcc` for some reason or the libs were also installed with Homebrew, but usually the OpenGL and OpenAL headers are under `OpenGL/` and `OpenAL/` respectively.

zeromus:
This wouldn't be a problem if you set autocrlf to false. Any other setting is universally invalid and results in these whitespace noise commits _at best_ and broken build scripts at worst. If you're wondering why the other settings exist if they're universally invalid, it's because they're trolls on windows users.

zeromus:
I agree, choosing autocrlf=false is important to preserve line breaks as they are.

theCapypara:
Touché I guess.
Imo git should really add a way to add settings like these directly to version control, so you can enforce things like this

theCapypara:
@tech-ticks @zeromus
Is there anything blocking this to be merged?

theCapypara:
Also interestingly, the current master seems to build to MacOS I just noticed.

rofl0r:
> Edit: It did not build before, though it probably still doesn't load as it currently stands unless we merge these changes.

i fixed the osx ci build recently independently. too me these changes here look quite suspicious, like changing random things until it worked, then doing git commit -am.
if it doesn't load, as you suppose, then this should likely be going into a new issue.
i'm in favor of closing this PR as "works for me".

theCapypara:
But from what I've gathered as-is the Mac version doesn't work. Or is that incorrect? I don't have a Mac to test myself, maybe it's just the instructions that need updating? What would be the proper solution to this?

zeromus:
gathered from what? rofl0r says it works. i gather that it works.

theCapypara:
@tech-ticks says it doesn't and a few months ago on my hackintosh I was also not able to get it to build.

rofl0r:
just click on the green checkmark on latest master commit and you'll see the CI built all targets successfully.

theCapypara:
Do they run though? I know they build but @tech-ticks reported they don't actually run.

zeromus:
I'm tired of reading about this. If someone who can actually build and run the osx build finds it's broken, we'll find out.

Issue #437, opened 2021-03-15
LINUX - The flatpak version one day wouldn't start with error: app.recent 72:expected end of input 


tvvoty:
The full text of the error:
(desmume:2): GLib-GIO-ERROR **: 19:40:35.931: g_menu_item_set_detailed_action: Detailed action name 'app.recent('file:///run/user/1000/doc/67bf012b/Subarashiki%20Kono%20Sekai%20-%20It's%20a%20Wonderful%20World%20(Japan)%20(Rev%201).nds')' has invalid format: 72:expected end of input

Fixed it with deleting the rom entry from /home/(user)/.var/app/org.desmume.DeSmuME/data/recently-used.xbel. 

I didn't find anything about it so I thought I should report it here in case anyone else has it.

linkmauve:
Hi, do you have a backup of that recently-used.xbel file before you had to remove that file, so that I could reproduce here?

tvvoty:
Yeah, it rewrites the entry in the file every time I load the game so I actually have to delete it every time, so I have the file. https://drive.google.com/file/d/1hPA4rD-kfaMRRM6_Sw27WNzrWtc8xbnr/view?usp=sharing . 

I guess it might be rom specific since it doesn't happen to other games? Should I post the rom as well?

Salz:
The issue is the single qoute (') in the filename, which does not get escaped by gtk.
https://github.com/TASEmulators/desmume/pull/491 has a fix for this.

It changes '%s' to "%s" because double quotes do get uri-escaped by gtk (as specified by RFC 2396).

linkmauve:
Fixed in #491.

PR #491, opened 2021-11-19
Linux: Fix recent entries with quote in filename

Salz:
According to RFC 2396, the single quote character (') is allowed in uri
strings and is not escaped by gtk, so the action string constructed for
the recent files menu must be quoted with " instead of '.

This fixes issue #437


linkmauve:
Good catch, thanks!

Issue #407, opened 2020-12-17
LINUX - Can't get menu bar back after disappearing


D0J0P:
I downloaded the latest version as a Flatpak. I accidentally clicked to hide the menu, and there's no way in the GUI to bring it back. I tried right-clicking everywhere, and the context menu doesn't show an option to bring the menu bar back. I tried uninstalling and reinstalling, but it reinstalled with the same settings.

linkmauve:
How do other software solve this issue?  Or did they drop the ability to hide the menu?  I’m not well-versed in UI design, sorry. :x

D0J0P:
I don't even think DeSmuMe on Windows 10 has a "hide menu" option, only a "hide toolbar" option, which is very easy to undo as the menu is still there. I can't do that on Linux as the entire menu is gone, and there's no way in the GUI to get it back.

zeromus:
brainstorming: hit hotkey to bring it back and write hotkey in manual, or write config key to manually edit in manual. Or don't save that setting and make users hide it every time.

tvvoty:
So, is there a way to get it back at all? I tried reinstalling it but that didn't work.

tvvoty:
> brainstorming: hit hotkey to bring it back and write hotkey in manual, or write config key to manually edit in manual. Or don't save that setting and make users hide it every time.

A hotkey makes the most sense. Atom uses Alt for showing\hiding the menu for example.

Salz:
The patch in https://github.com/TASEmulators/desmume/pull/492 brings back the pop-up menu from gtk2, which should fix this issue.

linkmauve:
Fixed in #492.

PR #492, opened 2021-11-19
Linux: Add pop-up menu back

Salz:
This adds the view menu back as pop-up menu, which was disabled by the
gtk2 port.

This fixes issue #407.


linkmauve:
Thanks for the fix!

As a process remark, if you reword the last line of your commit as “Fixes #issue-number”, it will automatically close the issue in question once I merge it.

Issue #415, opened 2021-01-08
LINUX - GTK build segfaults when choosing a slot1 type (like R4)


MaximumGeeker:
*Copied as is from desmume [forum topic](http://forums.desmume.org/viewtopic.php?pid=28843#p28843)*

I've found out that linux builds (don't know about windows or mac osx) seg fault with a call to 'slot1_Change' whenever a slot1 type different than the default RETAIL is passed with '--slot1'. With considerable effort (I should learn me some good gdb-fu) I've traced the issue to the function 'desmume_init()', which calls 'NDS_Init()' , being called after the 'process_addonCommands()' (a method of 'configured_features' class) in the main function of GTK linux build. CLI linux version does not have this issue, in fact I found out the issue by comparing GTK's code to CLI's.

Specifically, 'desmume_init()' provides for all slot1 types constructors to be defined, but in GTK code they are accessed before that with 'my_config->process_addonCommands()', hence the segfaults. I got it all worked out by moving 'desmume_init()' call above 'my_config->process_addonCommands()'.

Just wanted to mention it so it gets fixed in upstream, and there's another issue about all command line switches except '--help' not working in upstream, should I make another topic for it or what?

--mg

Salz:
The patch in https://github.com/TASEmulators/desmume/pull/493 moves NDS_Init() to the start of the program, leaving the sound settings in desmume_init() after parsing the command line options. This is similar to the desmume-cli frontend.

linkmauve:
Fixed in #493.

PR #493, opened 2021-11-19
Linux: Call NDS_Init() early

Salz:
Since NDS_Init() takes no arguments, it should not hurt to call it early
in the gtk frontend, too.

This fixes the segfault in issue #415


linkmauve:
Thanks for the fix!

PR #495, opened 2021-12-14
ci: Add macOS app Build

rinsuki:
This pull-request adds macOS app (means `DeSmuME.app`) build to GitHub Actions.

which makes easier for macOS users to try latest DeSmuME build.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   build:
-    name: Build DeSmuME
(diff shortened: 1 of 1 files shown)


zeromus:
I think rogerman likes to make his own builds, maybe due to PGO or something I don't know, but this is definitely an improvement

rofl0r:
since this PR was merged, another OSX CI build fails constantly: https://github.com/TASEmulators/desmume/runs/4526393503?check_suite_focus=true

PR #503, opened 2022-02-08
add github workflow to build win64 artifact

### LINKED ISSUE #500: Issue with AppVeyor
So not sure if this is just me or not. Never used AppVeyor before, but when I go to download the build it gives me this message...

{"message":"Artifacts download limit (1024 MB/day) exceeded."}

Is there something I'm doing wrong orrrrrrrrr?

rofl0r:
special thanks to @SimonAfek who figured out some nitty-gritty stuff
to make this work.

closes #500


rinsuki:
it might be good to add as job of build.yml instead of create build_win.yml.

zeromus:
Why?

rinsuki:
because:

- build.yml already including macOS artifacts. i think it will be good to add Windows artifacts to artifacts list that already including macOS artifacts.
- We (or someone else) may want to release automatically from GitHub Actions (in the future), but GitHub Actions can't download artifacts from other workflow file.

rofl0r:
superseded by #505

PR #517, opened 2022-03-30
Interface: Added a function to get/set the ARM9 next instruction

theCapypara:


zeromus:
did you test this with the jit enabled?

theCapypara:
I didn't; I will add a comment pointing out it's untested with JIT on for now.

theCapypara:
I changed it to return 0 for the getter and be a no-op for the setter. I can also change it to crash instead, but I feel like this is a bit less radical, given that this is supposed to be linked by another process. Whoever is calling is is responsible to make sure JIT is disabled (which as of right now there's no way to enable it from the interface anyway).

zeromus:
ok I would still rather it abort but whatever

PR #518, opened 2022-03-30
Interface: Fix memory leak on opening ROMs, added function to manually free ROMs

theCapypara:
As a side note, at a quick glance it seems that this also affects the Windows and Linux GUI frontends (didn't check Mac), since I couldn't find `NDS_FreeROM();` being called there when opening a new ROM. The Windows GUI only seems to free the old ROM when explicitly told to close the ROM. However this is only from a quick glance at the code, I didn't actually do more investigation, but it might be worth to check.

### DIFF
--- a/desmume/src/frontend/interface/interface.cpp
+++ b/desmume/src/frontend/interface/interface.cpp
@@ -39,6 +39,7 @@
 
 #define SCREENS_PIXEL_SIZE 98304
 volatile bool execute = false;
+volatile bool rom_opened = false;
 TieredRegion hooked_regions [HOOK_COUNT];
 std::map<unsigned int, memory_cb_fnc> hooks[HOOK_COUNT];
 
@@ -97,9 +98,21 @@ EXPORTED void desmume_set_language(u8 lang)
 EXPORTED int desmume_open(const char *filename)
 {
     int i;
+	if (rom_opened) {
+		NDS_FreeROM();
+	}
     clear_savestates();
     i = NDS_LoadROM(filename);
+	if (i > 0) {
+		rom_opened = true;
+	}
     return i;
+}
+
+EXPORTED void desmume_close()
+{
+	NDS_FreeROM();
+	rom_opened = false;
 }
 
 EXPORTED void desmume_set_savetype(int type) {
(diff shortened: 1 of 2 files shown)


desmume/src/frontend/interface/interface.cpp
```diff
@@ -97,11 +98,23 @@ EXPORTED void desmume_set_language(u8 lang)
 EXPORTED int desmume_open(const char *filename)
 {
     int i;
+	if (rom_opened) {
```
rofl0r, line 101: you seem to have used spaces for indent in the past, now tab

theCapypara: Thanks, I normalized it for those lines. Though the file still contains whitespaces differences, but I think this is a more general issue.

desmume/src/frontend/interface/interface.cpp
```diff
@@ -39,6 +39,7 @@
 
 #define SCREENS_PIXEL_SIZE 98304
 volatile bool execute = false;
+volatile bool rom_opened = false;
```
rofl0r, line 42: i don't think this needs to be volatile, unless EXPORTED can be declared as static. otherwise there's an interface barrier using the extern function and the compiler can't optimize out the access to the variable.

theCapypara: Brushed up my knowledge about volatile and static (I'm really not much of a C/C++ dev...) and yeah, I shouldn't just have copy-pasted that. I changed it to static.

desmume/src/frontend/interface/interface.cpp
```diff
@@ -97,11 +98,23 @@ EXPORTED void desmume_set_language(u8 lang)
 EXPORTED int desmume_open(const char *filename)
 {
     int i;
+    if (rom_opened) {
+        NDS_FreeROM();
+    }
     clear_savestates();
     i = NDS_LoadROM(filename);
+    if (i > 0) {
+        rom_opened = true;
```
rofl0r, line 107: if rom_opened was already true, it will stay true even if `i <= 0`

Issue #522, opened 2022-04-09
Can't find option for General\BackupSavestateSuppress in GUI's menus


Keppl:
While trying to figure out why SaveSlots folder keeps growing I found this commit: 0d65ea6050de2d47f62a8eb0148864ee6583d2fc
I understand the reason for it (though I'm not sure why it is necessary to keep every single one as oppose to just last <n> of them).

I see that there is a check for suppressing this behavior, by way of "[General] BackupSavestateSuppress=1", which works by editing desmume.ini file manually. 

On the other hand, I couldn't find option for it anywhere in GUI's menus. Unless I missed it... 

This is based on version complied from 2e6b9383

zeromus:
There's no option for it in the GUI menus. 

It isn't necessary to keep every single one of them. You're welcome to contribute a patch that keeps just the last of them.

Keppl:
Here is a quick stab at keeping 5 latest backups, rotating: [53be5f69687cbd1c23a6287c76ea81a7fbd400dd](https://github.com/Keppl/desmume/commit/53be5f69687cbd1c23a6287c76ea81a7fbd400dd)

Probably better to have some buffer in number of backup files in case someone presses load more than once by accident.

It was a  while since I wrote anything in C++, hopefully I didn't make a mistake somewhere. Seems to run but I out to test it bit more.

I can make a PR if this is ok?
(after I tested it bit more).

zeromus:
OK that was a pretty good idea to keep it simple, and the filesystem date can help find the most recent. But set the limit to like... 200. I don't think 5 is enough. Even though in most cases people are loading states by accident and realizing later it's a mistake, people will guess which state to load and try several (perhaps up to 10 because there's 10 slots, see?). I think 200 is enough margin for error.. I doubt anyone uses much more than that in one long sit-down and play session, which seems like a good target to me

Keppl:
Ok, I set default to 200. 
I also added check for [General]BackupSavesMax , so that that there would be a way for user to set max themself.
 (At least on windows, can't quite figure out linux/posix equivalent of GetPrivateProfileInt. It looks like process for posix config is bit different).
I left BackupSavestateSuppress to avoid breaking someones config, but BackupSavesMax=0 should be equivalent.
I will create a PR in a bit.

On overall approach, I guess I was looking it from slightly different angle. I though more along the lines of "whoops I pressed wrong button" which usually an immediate response.
[...]

PR #523, opened 2022-04-12
Limit maximum number of backups files for load state backup

Keppl:
Keep only maximum of <n> (200 by default) backups when creating backup during state load.
Once maximum files exist overwrite oldest one.

Also added ability to define <n> in desmume.ini under [General] BackupSavesMax=<n>

Based on discussion in https://github.com/TASEmulators/desmume/issues/522

### DIFF
--- a/desmume/src/saves.cpp
+++ b/desmume/src/saves.cpp
@@ -765,19 +765,24 @@ void loadstate_slot(int num)
 				index_file = fopen(index_fname.c_str(), "w"); // Create if doesn't exist
 			}
 			
-			cur_index = (cur_index + 1) % max_index; // next
+			if (index_file) {
+				cur_index = (cur_index + 1) % max_index; // next
 
-			fprintf(index_file, "%d", cur_index); // Store new index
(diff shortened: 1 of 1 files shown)


desmume/src/saves.cpp
```diff
@@ -744,25 +748,37 @@ void loadstate_slot(int num)
...
-				{
-					seed = rand()*16000+rand();
-					fclose(f);
-					continue;
-				}
-				seed++;
-				savestate_save(fname.c_str());
-				printf("Creating backup of current state prior to loadstate as path: %s\n",fname.c_str());
-				break;
+				index_file = fopen(index_fname.c_str(), "w"); // Create if doesn't exist
```
rofl0r, line 765: this may return 0 too for whatever reason, leading to a crash later on

Keppl: Good point.
I added check to make sure that index was opened, and if not skip backup part.
I'm assuming it is still preferable to proceed with loading state even if backup can not be created.

PR #535, opened 2022-05-28
Use -fileSystemRepresentation for NSStrings going to lower-level APIs.

MaddTheSane:


rogerman:
It seems like you're hell-bent on everyone's software on GitHub using [NSString fileSystemRepresentation], and so I'll play along.

I implemented your changes by calling these "fileSystemRepresentation" methods indirectly via CocoaDSUtil. The new methods will serve as syntactic sugar to denote what the exact intent of these methods are in their context.

I also removed your changes in cocoa_cheat.mm that would have broken cheats in the Cocoa port. Just because something calls [NSString cStringUsingEncoding:NSUTF8StringEncoding] doesn't mean it pertains to a file path -- please be more mindful of your copy/paste stuff before submitting a PR, thanks.

zeromus:
@MaddTheSane what exactly is the point of the crusade you're on?

MaddTheSane:
What do you mean?

zeromus:
rogerman made it sound like youre making similar PRs in other projects

MaddTheSane:
It's a fairly easy change.

zeromus:
why make this fairly easy change

rogerman:
It's okay, @zeromus. Using the [NSString fileSystemRepresentation] method technically is the proper form for dealing with C-function file paths in Cocoa. It doesn't hurt backwards compatibility, and so I should have done this from the very get-go. I screwed up, and @MaddTheSane corrected me on it. I was just miffed with the error in cocoa_cheat.mm, so I may have come off as a bit annoyed in my comment. That's all this is.

PR #539, opened 2022-05-30
Fix Keyboard Input bugs for Linux CLI and Interface

Lino5000:
This fixes two bugs related to keyboard input for the Linux CLI build, which also fixes them for the Interface build since they make use of the same input code.

1. Allow most of the keyboard input bindings to actually work, by changing the `keyboard_cfg` array and the `lookup_key` function to use `u32`s rather than `u16`s, aligning them with the other keyboard input code. This was a problem because the input configuration is set with a `memcpy` from the default configuration which is a `u32[]` into the `keyboard_cfg` array, resulting in a mangled configuration as extra bytes were added, later keys were missing, and some key codes that require 32 bits were split and therefore invalid.
2. Allow a `KEYDOWN` event of the Right Shift key to both trigger a modifier as it used to, and also act to press an emulated button. This is needed as the default configuration is for Right Shift to be Select, but the switch statement that handled input never let a Right Shift event pass through to the emulated inputs. It is fixed simply by allowing a fall-through of that case into the default.

### DIFF
--- a/desmume/src/frontend/posix/shared/ctrlssdl.cpp
+++ b/desmume/src/frontend/posix/shared/ctrlssdl.cpp
@@ -513,7 +513,6 @@ process_ctrls_event( SDL_Event& event,
                 break;
             case SDLK_RSHIFT:
                 shift_pressed |= 2;
-                break;
             default:
                 key = lookup_key(event.key.keysym.sym);
                 ADD_KEY( cfg->keypad, key );
--- a/desmume/src/frontend/posix/shared/ctrlssdl.cpp
+++ b/desmume/src/frontend/posix/shared/ctrlssdl.cpp
@@ -30,7 +30,7 @@
 
 
 
-u16 keyboard_cfg[NB_KEYS];
+u32 keyboard_cfg[NB_KEYS];
 u16 joypad_cfg[NB_KEYS];
 u16 nbr_joy;
 mouse_status mouse;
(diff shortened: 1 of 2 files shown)


desmume/src/frontend/posix/shared/ctrlssdl.cpp
```diff
@@ -513,7 +513,6 @@ process_ctrls_event( SDL_Event& event,
                 break;
             case SDLK_RSHIFT:
                 shift_pressed |= 2;
-                break;
```
rofl0r, line 516: should add a `/* fall-through */` comment here to mute compiler warnings

desmume/src/frontend/posix/shared/ctrlssdl.cpp
```diff
@@ -513,7 +513,7 @@ process_ctrls_event( SDL_Event& event,
                 break;
```
rofl0r, line 513: why is it ok for lshift to break, but not rightshift ? shouldn't both fall-through (or goto) the default case ?

rofl0r: in which case the simplest solution would be to factor the default case to outside the switch, so it's always executed.

rofl0r: the keyboard config goes through desmume_config_read_file so it should indeed be possible to set left shift as an action button too.

zeromus:
yes, take the ifs out of the switch and then nuke the switch entirely.

rofl0r:
merged as ba05516b...b2ad7224

PR #545, opened 2022-06-11
Add basic support for LoongArch  Linux

phorcys:
LoongArch is a new RISC  cpu architecture.
This patch add basic support for building and running under  LoongArch Linux.

### DIFF
--- a/desmume/src/libretro-common/features/features_cpu.c
+++ b/desmume/src/libretro-common/features/features_cpu.c
@@ -232,6 +232,8 @@ retro_time_t cpu_features_get_time_usec(void)
 
 #if defined(__x86_64__) || defined(__i386__) || defined(__i486__) || defined(__i686__)
 #define CPU_X86
+#elif defined(__loongarch__) || defined(__loongarch64)
+#define CPU_LOONGARCH
 #endif
 
 #if defined(_MSC_VER) && !defined(_XBOX)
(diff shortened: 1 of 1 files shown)


rofl0r:
having to explicitly specify code for each available architecture looks like a maintenance nightmare in the long run. can we find a solution that looks like `#if list_of_architectures_with_special_optimizations    special_code();    #else     generic_code();` ?

zeromus:
the only acceptable commit which modifies a file in libretro-common directory will come with a comment "updated from latest libretro common". if you want to support your cpu, youll have to support it over there first.

SpriteOvO:
> the only acceptable commit which modifies a file in libretro-common directory will come with a comment "updated from latest libretro common". if you want to support your cpu, youll have to support it over there first.

@zeromus Does "there" mean https://github.com/libretro/libretro-common? (I'm not sure, because there are significant inconsistencies in the content of the files). I am planning to open a PR to support compiling for RISC-V.

zeromus:
Yes. That they're diverged is unfortunate. That's why we aren't going to diverge them more. Good luck. It's possible if these are the only changes you really needed in libretro-common and it's minuscule compared to more serious changes elsewhere, that we'll merge these. But we'll have to wait and see.

SpriteOvO:
After that I opened a PR https://github.com/libretro/libretro-common/pull/200 over there, but it ended up being reverted because it broke builds on `MIPS32` (https://github.com/libretro/RetroArch/commit/6b9a382f2e90923b38fcca1f5a0d11ffe391e3fe).

Upon further investigation I found that the code that really broke builds for architectures other than `x86` and `ARM` wasn't in the upstream https://github.com/libretro/libretro-common, it is introduced from the commit 907cbf8308dc21153098806050bc58845919ae76, and this bug has been fixed in the commit d4afd4977ce8b7ba6293946c74914afe79536059.

Since a version containing that fix was not released, I did not notice it. So no action needed here right now, sorry for the noise.

PR #546, opened 2022-06-12
Enhanced interpolation, and fixed-point sample positions

Aikku93:
This branch was originally started as an attempt to add sinc interpolation, but it kinda blew up from there.

The main changes here are:

1. All `sampcnt` and `sampinc` are now .32 fixed-point values, stored in 64-bit integers. In theory, this shouldn't have been necessary, but it does simplify some things, and should avoid issues with floating-point drift.
2. New SPU savestate format. This was tested for backwards compatibility by loading savestates generated from the last master branch, and I haven't seen any issues.
3. PSG square waves are now generated programmatically, without a LUT. I don't think this makes much difference, but it definitely can't hurt.
4. All channels now read sample data into a circular buffer instead of reading directly from memory. This might pose a slight performance penalty for PCM waveforms (eg. with a large `sampinc`, it must now process all the samples in between instead of just jumping ahead), but ADPCM should be unaffected as it had to decode all samples in between anyway.
5. Some refactoring of the operations in SPU.cpp. This was to allow for easy integration of new interpolation methods.
6. Addition of Catmull-Rom spline-based interpolation. The spline equations were refactored to obtain a 4-tap FIR filter, and this is used as the interpolation kernel.

As an aside, I did end up implementing the sinc interpolation I started this whole journey for, but as it turns out, you need a rather obscene number of terms (upwards of 16) to get even moderately good results, even when the sinc function is windowed with a heavy-duty kernel like Blackman-Harris. At this point, I feel that adding a long sinc kernel would just be a heavy performance penalty more than anything, so I've left it out. If there's any interest in restoring it, however, the refactoring should make this simple enough.

---

With all that out of the way, I do have an issue: I only have one computer at home, and it's relatively powerful. So as much as I prefer the new interpolation method I've added, I'm not sure how much of a penalty my refactoring is incurring. So if this PR goes anywhere towards merging into master, I believe that it will require more testing to see just how bad things are.

### DIFF
--- a/desmume/src/SPU.cpp
+++ b/desmume/src/SPU.cpp
@@ -1096,11 +1096,6 @@ static FORCEINLINE s32 FetchADPCMData(channel_struct *chan, s32 pos)
 {
 	if(pos < 8) return 0;
 
-	const u32 shift    = (pos&1) * 4;
-	const u32 data4bit = ((u32)read08(chan->addr + (pos>>1))) >> shift;
-	const s32 diff = precalcdifftbl [chan->index][data4bit & 0xF];
-	chan->index    = precalcindextbl[chan->index][data4bit & 0x7];
-
 	s16 last = chan->pcm16b[SPUCHAN_PCM16B_AT(chan->pcm16bOffs)];
 
 	if(pos == (chan->loopstart<<3)) {
@@ -1108,6 +1103,11 @@ static FORCEINLINE s32 FetchADPCMData(channel_struct *chan, s32 pos)
 		chan->loop_pcm16b = last;
 		chan->loop_index = chan->index;
 	}
+	
+	const u32 shift    = (pos&1) * 4;
+	const u32 data4bit = ((u32)read08(chan->addr + (pos>>1))) >> shift;
+	const s32 diff = precalcdifftbl [chan->index][data4bit & 0xF];
+	chan->index    = precalcindextbl[chan->index][data4bit & 0x7];
 
 	return MinMax(last + diff, -0x8000, 0x7FFF);
 }
--- a/desmume/src/SPU.cpp
+++ b/desmume/src/SPU.cpp
@@ -126,6 +126,14 @@ static FORCEINLINE T MinMax(T val, T min, T max)
 	return val;
 }
 
+// T must be unsigned type
+template<typename T>
+static FORCEINLINE T AddAndReturnCarry(T *a, T b) {
+	T c = (*a >= (-b));
+	*a += b;
+	return c;
+}
+
 //--------------external spu interface---------------
 
 int SPU_ChangeSoundCore(int coreid, int newBufferSizeBytes)
@@ -380,7 +388,8 @@ static FORCEINLINE void adjust_channel_timer(channel_struct *chan)
 	//  ARM7_CLOCK / (DESMUME_SAMPLE_RATE*2) / (2^16 - Timer)
 	// = ARM7_CLOCK / (DESMUME_SAMPLE_RATE*2 * (2^16 - Timer))
 	// ... and then round up for good measure
-	chan->sampinc = ((u32)ARM7_CLOCK*(1ull<<32)-1) / (DESMUME_SAMPLE_RATE*2ull * (0x10000 - chan->timer)) + 1;
+	u64 sampinc = ((u32)ARM7_CLOCK*(1ull << 32) - 1) / (DESMUME_SAMPLE_RATE * 2ull * (0x10000 - chan->timer)) + 1;
+	chan->sampincInt = (u32)(sampinc >> 32), chan->sampincFrac = (u32)sampinc;
 }
 
 void SPU_struct::KeyProbe(int chan_num)
@@ -431,23 +440,23 @@ void SPU_struct::KeyOn(int channel)
 	case 0: // 8-bit
 	//	thischan.loopstart = thischan.loopstart << 2;
 	//	thischan.length = (thischan.length << 2) + thischan.loopstart;
-		thischan.sampcnt = -3 * (1ll<<32);
+		thischan.sampcntFrac = 0, thischan.sampcntInt = -3;
 		break;
 	case 1: // 16-bit
 	//	thischan.loopstart = thischan.loopstart << 1;
 	//	thischan.length = (thischan.length << 1) + thischan.loopstart;
-		thischan.sampcnt = -3 * (1ll<<32);
+		thischan.sampcntFrac = 0, thischan.sampcntInt = -3;
 		break;
 	case 2: // ADPCM
 		thischan.pcm16b[0] = (s16)read16(thischan.addr);
 		thischan.index = read08(thischan.addr + 2) & 0x7F;
-		thischan.sampcnt = -3 * (1ll<<32);
+		thischan.sampcntFrac = 0, thischan.sampcntInt = -3;
 		thischan.loop_index = K_ADPCM_LOOPING_RECOVERY_INDEX;
 	//	thischan.loopstart = thischan.loopstart << 3;
 	//	thischan.length = (thischan.length << 3) + thischan.loopstart;
 		break;
 	case 3: // PSG
-		thischan.sampcnt = -1 * (1ll<<32);
+		thischan.sampcntFrac = 0, thischan.sampcntInt = -1;
 		thischan.x = 0x7FFF;
 		break;
 	default: break;
@@ -756,7 +765,7 @@ void SPU_struct::ProbeCapture(int which)
 	u32 len = cap.len;
 	if(len==0) len=1;
 	cap.runtime.maxdad = cap.dad + len*4;
-	cap.runtime.sampcnt = 0;
+	cap.runtime.sampcntFrac = cap.runtime.sampcntInt = 0;
 	cap.runtime.fifo.reset();
 }
 
@@ -1163,7 +1172,7 @@ static FORCEINLINE void MixLR(SPU_struct* SPU, channel_struct *chan, s32 data)
 template<int FORMAT> static FORCEINLINE void TestForLoop(SPU_struct *SPU, channel_struct *chan)
 {
 	// Do nothing if we haven't reached the end
-	if((chan->sampcnt >> 32) < chan->totlength_shifted) return;
+	if(chan->sampcntInt < chan->totlength_shifted) return;
 
 	// Kill the channel if we don't repeat
 	if(chan->repeat != 1)
@@ -1196,8 +1205,8 @@ template<int FORMAT> static FORCEINLINE void TestForLoop(SPU_struct *SPU, channe
 	}
 
 	// Wrap sampcnt
-	s64 step = chan->totlength_shifted - (chan->loopstart << format_shift[FORMAT]);
-	while ((chan->sampcnt >> 32) >= chan->totlength_shifted) chan->sampcnt -= step * (1ll << 32);
+	u32 step = chan->totlength_shifted - (chan->loopstart << format_shift[FORMAT]);
+	while (chan->sampcntInt >= chan->totlength_shifted) chan->sampcntInt -= step;
 }
 
 template<int CHANNELS> FORCEINLINE static void SPU_Mix(SPU_struct* SPU, channel_struct *chan, s32 data)
@@ -1220,14 +1229,11 @@ template<int FORMAT, SPUInterpolationMode INTERPOLATE_MODE, int CHANNELS>
 	{
 		// Advance sampcnt one sample at a time. This is
 		// needed to keep pcm16b[] filled for interpolation.
-		// We need to do some janky things here to keep the
-		// fractional bits in place when we loop :/
-		s64 newsampcnt = chan->sampcnt + chan->sampinc;
-		u32 nSamplesToSkip = (u32)((newsampcnt >> 32) - (chan->sampcnt >> 32));
+		u32 nSamplesToSkip = chan->sampincInt + AddAndReturnCarry(&chan->sampcntFrac, chan->sampincFrac);
 		while(nSamplesToSkip--)
 		{
 			s16 data = 0;
-			s32 pos = chan->sampcnt >> 32;
+			s32 pos = chan->sampcntInt;
 			switch(FORMAT)
 			{
 				case 0: data = Fetch8BitData (chan, pos); break;
@@ -1239,14 +1245,13 @@ template<int FORMAT, SPUInterpolationMode INTERPOLATE_MODE, int CHANNELS>
 			chan->pcm16bOffs++;
 			chan->pcm16b[SPUCHAN_PCM16B_AT(chan->pcm16bOffs)] = data;
 
-			chan->sampcnt += 1ll << 32;
+			chan->sampcntInt++;
 			if (FORMAT != 3) TestForLoop<FORMAT>(SPU, chan);
 		}
-		chan->sampcnt = ((chan->sampcnt >> 32) << 32) | (u32)newsampcnt;
 
 		if(CHANNELS != -1)
 		{
-			s32 data = Interpolate<INTERPOLATE_MODE>(chan->pcm16b, chan->pcm16bOffs, (u32)chan->sampcnt);
+			s32 data = Interpolate<INTERPOLATE_MODE>(chan->pcm16b, chan->pcm16bOffs, chan->sampcntFrac);
 			SPU_Mix<CHANNELS>(SPU, chan, data);
 		}
 	}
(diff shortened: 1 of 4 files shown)


zeromus:
Let me know when you're done making commits...

Aikku93:
Yeah, sorry about that. I was confident I'd ironed out all the problems at each commit. :/

That should be the last of it, though. I've been testing everything I have in my library this time (instead of just a few easy-to-check samples), and haven't had any more issues show up.

zeromus:
What library are you making?

Aikku93:
Sorry, I meant I thoroughly tested the new code with my library/collection of DS programs, not that I'm working on a software library or anything like that.

zeromus:
seems FASTER, to me.
I have pretty good ears for debugging sound problems and use studio monitors at my desk, but I've never really been able to hear the difference between linear filtering and anything else in pretty much any context except possibly ones designed to be testcases, so I hope you get some pleasure yourself from the results since I'm not sure anyone else will. But thanks for making it faster, anyway.

Aikku93:
Huh, interesting. I honestly didn't expect it to be any faster by any margin; at most, I expected same performance. Maybe all those floating-point truncations were causing issues?

And yeah, for "general audio", linear interpolation is generally good enough. But when you get to really low rates (eg. short sinusoid loops and the like), it gets kinda annoying when the aliasing artifacts weren't part of the intended style (like they are in chiptunes, for example). Cosine interpolation sounds almost as irritating as no interpolation on general audio, to me personally, but I guess most people are fine with it (and it does sound closer to actual hardware at 44.1kHz). Catmull-Rom keeps "smooth" waveforms smooth at low rates, and keeps "noisy" waveforms nice and crisp, which is why I like it. A good way to test it is with chiptunes and noisy audio; chiptunes will sound really smooth (which isn't fitting, but sounds correctly bandlimited), whereas the noisy stuff aliases to higher frequencies, enhancing the perceived "brightness" of the sound.

rogerman:
In preparation for adding this feature to the Cocoa port, I've done my own testing of the new Catmull-Rom interpolation method and also reviewed the existing interpolation methods as a point of comparison. For my testing, I used Mackie HR824 studio monitors, as well as Beyerdynamic DT 770 Pro 250 ohm studio headphones.

My DeSmuME settings for the entirety of testing are as follows:
    Audio Engine - macOS Core Audio
    Advanced SPU Logic - ON
    Synchronization Method - P

But before we get to listening tests, let's talk about performance: Whatever interpolation method is used, the performance difference is negligible. The SPU interpolation barely even registers on my profiler, so if users are looking to gain FPS by switching interpolation methods, well... they won't.

So really, all interpolation methods do is change the way things sound, and that is what the bulk of this discussion will be. Let's start with music from synthesized sounds.

**Synthesized Music from "Phoenix Wright: Trials and Tribulations":**
Many games ported from the GBA use synthesized music consisting of various sine waves, square waves, triangle waves, and noise generators. For this kind of music, the harmonics that are generated are essential for things to sound correct. For this test, I'm listening to the "Phoenix Wright: Trials and Tribulations" OST, focusing on the "Trial" theme, as well as the opening sequence of "Turnabout Memories".

_No Interpolation_: Far and away the best choice for synthesized music. The entire "Phoenix Wright: Trials and Tribulations" OST sounds rich and full, just as Capcom intended. Because the OST is synthesized, it doesn't sound noisy in the slightest. The opening phrase of the "Trial" theme sounds fantastic in how it blends the harmonics together in such a way that the music sounds full, making the upcoming trial feel more weighty. The two ringing bell sounds in the opening sequence of "Turnabout Memories" are front and center, cutting through the mix perfectly, hinting at the ominous murder scene to follow. When the first electrocution sound occurs, the sound is very loud, noisy, and dirty, immediately shocking the viewer and transitioning them into the following scene. Excellent!

_Linear_: Since this interpolation method eliminates all of the harmonics, this interpolation method completely destroys this game's entire OST. The OST sounds completely flat, and the two ringing bells in the opening sequence of "Turnabout Memories" lose their punch and fade to the back of the mix. For the first electrocution sound, interpolating away the noise makes it sound less harsh, which isn't as attention-getting. Linear Interpolation makes this game sound so bad that the overall feel of the game is negatively affected. The worst option by far.

_Cosine_: Harmonics are only somewhat preserved, which makes this game's OST sound bad, but it isn't nearly as striking as Linear. An important plus for Cosine is that it does preserve the presence of the two ringing bells in the "Turnabout Memories" opening sequence, and it does retain enough harmonics and noise so that certain sounds can remain crisp and bright, such as the first electrocution sound. But overall, I wouldn't recommend Cosine Interpolation for synthesized music, as the loss of harmonics causes the fullness of the music to be lost.

_Catmull-Rom_: This sounds just like Linear, and so Catmull-Rom is still catastrophic to the synthesized music in this game's OST. Catmull-Rom does manage to preserve a little bit more highs, but that's about it. Catmull-Rom still suffers the same weaknesses as Linear, causing the entire OST to sound flat and dull, with sounds having poor presence in the mix.

Synthesized Music Ranking (out of 10, with 10 being the best):
10 - Perfect: No Interpolation
4 - Subpar: Cosine
2 - Marginal: Catmull-Rom
1 - Terrible: Linear

**Sequenced Music from "Advance Wars: Days of Ruin":**
Most NDS games use music that is sequenced from instrument samples. For this kind of music, the limited storage space of an NDS ROM combined with the low quality of the NDS sound hardware creates for many situations where the mixer generates a lot of suboptimal samples, resulting in extraneous noise and hissing. This effect is usually annoying to listen to, but not always. For headphone users, the unwanted harmonics, noise, and hissing could be a huge detriment to the gaming experience. For this test, I'm listening to the "Advance Wars: Days of Ruin" OST, which features a wide variety of music with very strong drive.

_No Interpolation_: Without any interpolation, this game's OST is pretty hit-and-miss. Waylon's Theme is a very gritty metal track that uses the extra noise to its advantage, giving it that much more presence and punch. Lin's Theme is another metal track that also benefits from the extra noise for presence and punch, especially on the lead guitar. Isabella's Theme, a pop track with orchestral overtones, is a mixed bag, in that the lead instrument, bass, and kick all maintain a powerful punch, but a noisy background hiss distracts from these good elements. A similar thing can be said with Gage's Theme, a fusion orchestral track, in which the individual instruments all sound powerful and require the additional harmonics to sound correct, but are also brought down by other unwanted harmonics. Unfortunately, Tasha's Theme, a thrash metal track, maintains the noisy hiss throughout the entire song, which basically kills the song for me. Tabitha's Theme, a drum-heavy pop track that otherwise benefits from the punchy drums, suffers from an annoying high-pitched harmonic throughout the song. Forsycthe's Theme, an orchestral track, suffers the most from both unwanted harmonics and noisy hiss, utterly destroying the track completely. Overall, I would say that the "Advance Wars: Days of Ruin" OST's inclusion of metal tracks helps to hide the flaws of mixing without interpolation, but most other NDS games will want to use some kind of interpolation to clean up the noise and the hiss in order to truly enjoy their OSTs.

_Linear_: Sequenced music really needs interpolation to clean it up, and Linear Interpolation really does a good job at that. However, it's almost too good at cleaning out frequencies in that highs continue to be interpolated away too aggressively. This makes cymbal hits sound like trash, making the main menu music sound bad. Unfortunately, this effect also kills Will's Theme, which relies on cymbal hits to transition from one part of the song to the next. Tasha's Theme is also completely destroyed as the song's cymbal thrashing is gimped into oblivion. Several songs become weaker without their punchiness, but Forscythe's Theme, a track that does not rely on punchy instruments, benefits greatly from Linear Interpolation, although the horns do lose their presence. Greyfield's Theme, another orchestral track, doesn't get off as lucky since it is a horn-heavy track that needs those harmonics to sound correct. Just like running No Interpolation, Linear Interpolation is also pretty hit-and-miss, where the tradeoff here is eliminating the extraneous harmonics and noise at the cost of having to listen to dull and uninspired sound. However, due to sequenced music becoming inherently dirty when mixed, one cannot underestimate how much of a sound quality improvement can result from even overly aggressive interpolation.

_Cosine_: Sequenced music truly shines with Cosine Interpolation. This interpolation method cleans up most of the unwanted harmonics and noise, while preserving just enough harmonics so that the instruments can actually sing. Penny's Theme, a techno track, sounds excellent, especially in the presto string and harpsichord sections of the song. The orchestral Forsycthe's Theme truly becomes an epic with clean strings and strong horn presence, while Gage's Theme becomes godlike in how Cosine Interpolation brings the violin parts forward while not being taken down by unwanted background harmonics. Tasha's Theme is completely fixed, where the interpolation cleanup allows the cymbals to run wild and the lead guitar to chug properly. Lin's Theme sounds perfect, where the lead guitar takes center stage and the kick drum truly stands out. Even Brenner's Theme, a great rocking metal track that suffers from lousy in-game instrument samples, is brought up to something sounding decent due to Cosine Interpolation cleaning up the noise just enough for the lead guitar to come through the mix. However, Cosine Interpolation is not perfect under all conditions. For example, Waylon's Theme actually sounds a bit worse with Cosine Interpolation than No Interpolation since the interpolation process cleans up the music too much. For all these reasons, and since sequenced music is the most common type of music for NDS games, Cosine Interpolation is DeSmuME's default option.

_Catmull-Rom_: Again, Catmull-Rom Interpolation sounds very very similar to Linear Interpolation, and thus suffers from all of the same weaknesses. And yet again, Catmull-Rom does maintain better highs than Linear, which makes the cymbal hits in the main menu music, Will's Theme, and Tasha's Theme not suck as much. But notably, Forscythe's Theme, which benefits most from Linear Interpolation, actually sounds halfway decent with Catmull-Rom since the highs don't completely disappear from the strings and horns, thus giving these instruments at least some presence in the mix. Overall, Catmull-Interpolation sees some marginal improvements over Linear Interpolation, but it simply can't touch Cosine Interpolation when it comes to sequenced music.

Sampled Instrument Music Ranking (out of 10, with 10 being the best):
9 - Excellent: Cosine
6 - Good: Catmull-Rom
5 - Average: Linear
2 - Marginal: No Interpolation

**Vocal Music from the theme song of "Dengeki RPG: Cross of Venus Special", "Odyssey" by Yuuka Nanri:**
Many games use anime openings along with streamed music that plays directly off the NDS ROM. "Dengeki RPG: Cross of Venus" is a JRPG that does just that. This particular song was chosen for testing since there is a studio release version of the song, "Odyssey" by Yuuka Nanri, that can be compared against while doing changes to the interpolation method. Note that it seems like the game version is a rough cut of "Odyssey", while the studio release is a final cut, as noted by subtle differences in the song's EQ'ing and volume levels between the two versions, so there won't be a 1:1 comparison here. However, there should be enough similarities between the game version and the studio version so that we can at least have a rough idea of what Yuuka Nanri's vocal intonations, inflections, and resonance points are to still make an accurate comparison.

_No Interpolation_: This is the best option for vocal music, and it also avoids a big problem here in that extraneous harmonics as a result of NDS hardware audio mixing don't come into play here since we're just playing back an audio sample file. On the other hand, all of the interpolation methods lose too much sound detail, and so certain subtle instrumentals are lost in interpolation, such as the quiet strumming of the clean guitar, but are preserved when running No Interpolation. In addition, Yuuka Nanri's voice as a soprano singing a mezzo song is perfectly preserved, so I can immediately hear that it is, in fact, her singing this song. However, noise and aliasing issues (due to low bit rates and low sampling rates of the NDS audio file) continue to clutter up the mix, making it sound dirty compared to the studio version, so some users may not appreciate that aspect.

_Linear_: Yup, Linear Interpolation cleans up the noise real good, but it is still by far the worst option here. The background singers lose their intelligibility, the instrumental balance is completely off (especially the clean guitar, which very nearly disappears), and any EQ'ing that might have went into this song sounds flat. But worst of all, because all of the highs are interpolated away, Yuuka Nanri's voice is so thoroughly destroyed to the point where she sounds like a completely different singer altogether. It's extremely bad. If this were a real mixing job, then Linear Interpolation would change the vocalist's voice enough where there would be a guaranteed loss of business. I don't think it's so bad that it would result in a lawsuit, but this is approaching that point. It would definitely be a big reputation hit.

_Cosine_: Cosine Interpolation is also pretty bad, but there is a subtle enough preservation in the highs that at least a small amount of Yuuka Nanri's vocal intonations and resonance points are preserved. With that said, one could still make a debatable argument that it is still Yuuka Nanri singing "Odyssey" here. Although Cosine doesn't clean up the sound as much as Linear does, Cosine does allow the background singers to maintain a small amount of presence in the mix, and the instrumental sound quality is okay enough where no single instrument disappears into a muddy mix. Therefore, while Cosine Interpolation isn't great, at least it has enough going for it that it isn't completely useless.

_Catmull-Rom_: Interestingly enough, Catmull-Rom seems to preserve highs better than Cosine, and so this is the best of the interpolation methods. Like Linear Interpolation, Catmull-Rom is still too aggressive at cleaning up the sound, but its preservation of the highs is enough where one can easily hear Yuuka Nanri's vocal intonations, and even some of her inflections, that one can plausibly identify that she is singing "Odyssey" here. Background vocals and instruments are clear enough to make out too, and the interpolation cleanup helps a lot here, even more so than Cosine. Overall, Catrmull-Rom is still bad, but it is certainly the best of the worst.

Vocal Music Ranking (out of 10, with 10 being the best):
7 - Very Good: No Interpolation
4 - Subpar: Catmull-Rom
3 - Passable: Cosine
0 - Horrific: Linear

**Conclusions:**
Is Catmull-Rom Interpolation actually useful? I would say yes! Catmull-Rom is very similar to Linear in the way it sounds, but it clearly preserves highs much better than Linear. In fact, Catmull-Rom could be a viable replacement for Linear if it ever came down to it. However, Catmull-Rom also suffers many of Linear's weaknesses in that both interpolation methods are too aggressive in cleaning up the sound, resulting in much sound detail loss. This isn't a big issue for sequenced music or vocal music, but completely it does seriously hurt the tone of synthesized music.

**_My findings:_**
**_Does Catmull-Rom sound smooth?_** _Yes, it does._
**_Does Catmull-Rom keep noisy waveforms nice and crisp?_** _No, it does not. It sounds similar to Linear, in that Catmull-Rom aggressively interpolates any and all noise away._
**_Do chiptunes sound really smooth?_** _Unfortunately, yes, which ruins the way the chiptunes sound._
**_Does Catmull-Rom sound bright?_** _In general, no, but there could be an argument for answering yes if we're talking about vocal music._

And although I've ragged on Linear Interpolation this whole time, it can still have benefits for people with low quality speakers, such as the ones built into most laptops and most computer monitors. While Linear does knock out the highs in the audio mix, this can be beneficial for low quality speakers, which tend to have tinny inaccurate treble, overly strong mids, and grossly underpowered bass. It is only when you use professional studio monitors like my Mackie HR824s that you will start to hear the difference, and even then, most users won't really care too much regardless.

Cosine Interpolation still retains its crown as the default option, being at least passable in all areas while also having the best sound quality for sequenced music, which is the most common mixing case for the NDS.

And of course, there's always running No Interpolation, which can upset headphone users with its generation of unwanted harmonics, noise, aliasing, and whatnot. However, it is the best way to listen to vocal music and is the only way to truly listen to synthesized music. It's too bad that there aren't any ways to have two separate interpolation options: one for synthesized sounds and another for sampled sounds. That way, the user can turn off interpolation for the synthesized stuff and run the interpolation method of their choice for the sampled stuff, thus getting the best of both worlds.

Aikku93:
Thanks for those very detailed comparisons.

In general, I agree with your findings (especially in regards to "chiptune" music, since that style is /intended/ to sound the way it does, and the bandlimiting is just a side-effect of resampling from the lower sampling rate chosen for the samples).

That said (and I'm not arguing for changing the default interpolation method at all), I'm surprised that "cosine" interpolation is generally outperforming the others, since all it's doing is just adding nonlinear behaviour to the linear interpolation kernel. In my (admittedly less-detailed) testing with the same headphones, the aliasing distortion was a massive deal-breaker for both "no interpolation" and "cosine", as I could hear the "hiss" without even trying to listen for it. Though that might just be because I've been working with audio DSP for so long that defects like those stand out like beacons for me, personally.

I think that the biggest problem in trying to find a "one-size-fits-all" interpolation kernel is that some studios worked with the lack of hardware interpolation (eg. getting streams running at 32kHz, intentionally adding high-freq noise to mask the aliasing and/or relying on how hardware sounds to get a certain "feel" in their music, etc.), others worked around it (eg. resampling to ridiculous rates and lowpass-filtering so that the aliasing is inaudible), and others still just decided that it didn't matter and just went with whatever they had on hand without paying much attention to how it performs on hardware.

In general, I think that None/Cosine are good for approaching the sound of hardware (and the games that work with it), Linear is okay for "general purpose" but can be a bit "dull" and lifeless, while Catmull-Rom is more of a slightly crisper version of Linear that performs even better when given a sound source with enough "noise" in the last octave of the host sampling rate (ie. 11kHz~22kHz @ 44.1kHz; applies very well for cymbals and supersaw synths in particular).

zeromus:
Maybe you can make a hashed sample database and tag them with some kind of very fine label (chip, sampled instrument, vocals) and then create a mapping from label to interpolation method. We still have to figure out what to do about streamed music, but in 99% of those cases (e.g. not the korg instrument or playback of mic recordings, e.g. nintendogs or mic testers maybe?) the first N nonzero "blocks" will be uniquely identifying, and you might could catch that case (sampledata at a given address changing without the voice stopping) and turn it into a hash of the first N blocks. I would tag those as streamed AND ALSO another tag (I mean, have the ability to have multiple tags) because the streamed stuff will have common characteristics (a lossy codec). Streamed stuff may also be delay/reverb or surround sound mixing buffers, which is another problem to deal with. hey, I'm just brainstorming.

Issue #553, opened 2022-06-19
Use system tinyxml


musuruan:
Hi. I was updating the RPM Fusion package. 

Fedora has a policy to unbundle system libraries:
https://docs.fedoraproject.org/en-US/packaging-guidelines/#bundling

Therefore I made a patch to use the system tinyxml library:
https://pkgs.rpmfusion.org/cgit/free/desmume.git/tree/desmume-0.9.13-tinyxml.patch

You may want to include this patch upstream.

rofl0r:
please open pull requests instead of linking to patches elsewhere, so they can be properly reviewed and commented on

musuruan:
Opened #554.

PR #554, opened 2022-06-20
Use system tinyxml

musuruan:
Use the system tinyxml library instead of the bundled one.


musuruan:
AFAIK tinyxml is available for a lot of system:
https://repology.org/project/tinyxml/versions
https://repology.org/project/libtinyxml/versions

Anyway it is absolutely fine for you to continue to bundle. In fact in #553 I stated "You *may* want to include this patch upstream". I was then asked for a PR.

So why did you ask for a PR if you - the desmume developers - weren't interested?

zeromus:
Someone else asked for the PR. And neither of us realized it would break the build. The PR is fine as long as it just changes the linux build configuration.

PR #562, opened 2022-07-10
SPU: Logic re-write, remove `spu-advanced` toggle

Aikku93:
This PR covers an almost complete re-write of all the SPU mixing logic (plus some miscellaneous bits and pieces related to it), which has been tested to achieve performance on-par with the "non-advanced" SPU logic from before my interpolation-related re-write.

Further improvements include:
- Much better performance when using the dummy SPU
- Fixing a bug I introduced in my interpolation re-write (ADPCM channels weren't restored correctly from savestates)
- Fixing another bug I introduced where using capture-based echo effects would break in dual-SPU mode
- Perfect (cycle-accurate) synchronization with regards to the host sampling rate
- Miscellaneous tweaks that resulted in better generated assembly and/or performance
- Fix a long-standing edge-case bug related to capture (when switching to the dummy SPU core and setting dual-SPU mode simultaneously, the capture units were not generating silence, which would cause noise when switching from the dummy SPU core)
- Fix an extremely obscure bug/oversight (which has probably never occurred) where capture units in one-shot mode would continue looping until manually disabled.
- Capture is now fully emulated when using the dummy SPU core (this can be disabled via the `ENABLE_DUMMY_SPU_CAPTURE` toggle in SPU.cpp, in which case capture will only generate silence)
- Overhead reduction when not mixing by increasing latency (eg. when updating the core SPU in dual-SPU mode, or using the dummy SPU core). This technically results in some loss of synchronicity for those cases, but should really only affect channel allocation and/or testing for channels to have ended.

I've tried my best to remove all references to the "advanced SPU logic" toggle, but have only really been able to test on Windows at the moment, so other platforms might not be functional at the moment (I believe the macOS build in particular needs some more attention).

Special thanks to Rise of Seren for better performance profiling than I could get on my computer.


rofl0r:
> - Fixing a bug I introduced in my interpolation re-write (ADPCM channels weren't restored correctly from savestates)
> - Fixing another bug I introduced where using capture-based echo effects would break in dual-SPU mode

it would be helpful if those 2 fixes were in separate commits, so someone reading git log can figure out the relation

Aikku93:
Alright, no problem. I'll attempt a rebase (apologies in advance if I mess something up; I'm still not all that comfortable invoking `git` directly).

Aikku93:
Before I mark this PR ready for review, I've got a couple of questions (mostly because I'm still trying to get used to contributing on projects instead of just working on my own ones).

So one of the changes was "Add support for Catmull-Rom interpolation on Linux (macOS still doesn't have the option)" ([b6da15a](https://github.com/TASEmulators/desmume/pull/562/commits/b6da15a0482e718574bd622a12f3aaae31db97ea)). Should this be split out into a separate "Add Catmull-Rom support for Linux/macOS" PR/issue (and hopefully have someone look over what needs to be done to incorporate it properly), or is it fine to keep it here? Not having that interpolation option won't break anything, so I figure it might be better to do it as a new PR.

My other question was: I partially removed the advanced SPU option from the macOS code, but I know there's more places that I've missed. If the toggle is left in, then even though it won't actually do anything, the code won't break. Should I undo those changes and leave it for someone who actually knows what they're doing (or possibly even just undo the changes and raise it as an issue to be addressed at another time)?

zeromus:
For this project our policy is break it and let someone else fix it. We can't wait for everyone to do everything perfectly or else nobody would ever do anything.

rogerman:
Just do it -- I'll fix it later.

@Aikku93: In the future, if you are working on unfamiliar code, then whenever you want to remove options and you see the flag variables controlled by accessor functions, it is better coding practice to hard code the getter function to YES/true or NO/false, and then comment out the flag variable assignment in the setter function. By doing this, you risk breaking absolutely nothing while still achieving your intent in removing the option.

Aikku93:
Alright, thanks for the help and tips. Appreciate it.

rogerman:
@Aikku93: I'm currently reviewing the addition of the Catmull-Rom interpolation method and trying to get a handle on what it can do from the user's perspective so that I can write a proper tooltip for it. What's probably going to happen is that the Cocoa port's GUI is going to be updated to add Catmull-Rom at the same time as the Advanced SPU Logic option is removed.

Aikku93:
Hm, before this version ends up in `master`, would it serve any purpose to store `SPUCHAN_PCM16B_SIZE` and `SPUCAPTURE_FIFO_SIZE` as part of the SPU savestate? That way, if those constants ever change in a different version, the savestate version won't need to change (as well as avoiding the need to include even more code in the savestate loading routine, since it's already getting a bit unwieldy from my changes).

Aikku93:
I've split up the single mega-commit into multiple ones, which should hopefully result in a clearer commit history. I've also more-or-less followed rogerman's advice re. the Cocoa port, and instead of removing all the associated code that I'm unfamiliar with, I've just commented out the line that actually writes to `CommonSettings.spu_advanced`.

Please let me know if there's still more commit refactoring to do. I think I'm starting to get the idea of how commits and PRs are supposed to work, but I'm still getting there.

PR #564, opened 2022-07-19
Fix use of internal storage of temporary buffer

janisozaur:
The entry index gets converted to a (temporary) string, from which a
pointer to internal data is taken and only consumed outside the loop,
where parent variable no longer exists.


janisozaur:
> First-time contributors need a maintainer to approve running workflows

This PR is still waiting

PR #566, opened 2022-07-20
Fix #563: mismatched buffer sizes for memcpy

### LINKED ISSUE #563: Possible mismatch of buffer sizes
A recent commit, 677c6bb90754d00842, changed how buffers are handled, but there is still possibility of a bug due to mismatched sizes passed to `memcpy`:

https://github.com/TASEmulators/desmume/blob/677c6bb90754d008428005e3d31a038a78c79695/desmume/src/frontend/posix/gtk/main.cpp#L1943

and later in that same function:

https://github.com/TASEmulators/desmume/blob/677c6bb90754d008428005e3d31a038a78c79695/desmume/src/frontend/posix/gtk/main.cpp#L1964

Can you please confirm?

janisozaur:


rofl0r:
this doesnt solve the underlying problem, the mismatch of int32 vs int16 in those both configs...

zeromus:
ok i guess we still have to fix it better but an improvement to not have a copy/paste error on the memcpy size is still an improvement.

PR #568, opened 2022-07-21
Drop superfluous check for negative value of unsigned types

janisozaur:


zeromus:
I dont understand the significance of > sometimes and >= others.
This code is gross. Unsigned numbers are horrible. Can you please change the parameters to be ints instead?

Aikku93:
From what I can tell, `srcPos >= list.size()` makes sure that the source is somewhere /inside/ the list (as expected), while `dstPos > list.size()` uses > instead of >= because an item can be moved to the end of the list (ie. `srcPos == list.size()`).

As for signedness: While I'd argue that something's gone /seriously/ wrong if either of those values are out of bounds (or even negative) by the time `move()` is invoked, I think it really would be better to use `int` instead. At least on Windows (I can't seem to find references to `move()` on other platforms), `s32 draggedItem` and `s32 highlightedItem` are the values passed to the function, both of which are signed values that store -1 when in an 'invalid' state.

rogerman:
This issue was fixed as part of some general code cleanup in commit aa21454. The offending line was deleted in the exact same way this PR did it.

PR #628, opened 2023-01-30
Fix a few compiler warnings

radioactiveman:
warning: use of bitwise '&' with boolean operands [-Wbitwise-instead-of-logical]

Reported by clang 15.


radioactiveman:
Upstream fix of my commit for https://github.com/audacious-media-player/audacious-plugins/commit/0329930e1eb5ab370d58d3ecaaa86b0b391b880a.

Interestingly `OP_UMULL_S` uses `&&` already, so this also improves consistency.

https://github.com/TASEmulators/desmume/blob/cfe88ce0d79d2d6d58799cfc20edd85ebc0ed02c/desmume/src/arm_instructions.cpp#L2405

zeromus:
thanks for finding "interestingly", else I would have doubted things, because the semantics of implicit conversions with bools are impossible to remember and I Would suspect we may actually be relying on a bug

PR #662, opened 2023-05-10
Windows CI: Update deprecated github actions

SimonAfek:
Fixes this warning about deprecated NodeJS versions in the GitHub-supplied actions `checkout` and `upload-artifact`:

image

Sample run [here](https://github.com/SimonAfek/desmume/actions/runs/4936859243).

PR #666, opened 2023-05-19
Update deprecated GitHub actions v2

SimonAfek:
As requested here https://github.com/TASEmulators/desmume/pull/662#issuecomment-1542165571

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,7 @@ jobs:
 
     steps:
       - name: checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: install dependencies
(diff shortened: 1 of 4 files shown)


SimonAfek:
Semi-related, but the mac build in `build_interface.yml` is failing because the `mac-10.15` image isn't available anymore (lowest one currently available is `mac-11`), so I'll do a separate pull request for that.

zeromus:
I dont know who's supposed to approve this. If it's DONE, let me know and I'll merge it

rofl0r:
it's correct and i merged it via squash in order to prevent someone from clicking the "merge" button which would lead to above ugly merge commits being added to git history which would make the history non-linear and result in potentially broken behaviour for git bisect.

PR #720, opened 2023-08-05
Fix potential index out of bounds

mjunix:


zeromus:
can you fix in slot2_Change() too

zeromus:
thanks

PR #737, opened 2023-10-10
Slot2: Add Slide Controller

windwakr:
This adds the Slide Controller add-on, used by the game Slide Adventure MAGKID. AFAIK no other emulator supports this add-on.
Only the Windows frontend is implemented. Set slot 2 to auto, load up the game and move your mouse around.

https://github.com/TASEmulators/desmume/assets/284886/6257d747-b54b-44f7-a05f-12309428a020

This is a draft because
- I'm not happy with my serial read/write code
- I'm unsure about my changes to main.cpp regarding raw input reading
- The cursor needs some mechanism to lock/hide it while playing

I'm not sure if I'll work on this again soon. If anyone else wants to then feel free to take over. Otherwise, it'll be here for anyone that just wants to play the game.

I mostly referenced the small section in [GBATEK](https://problemkaputt.de/gbatek.htm#dscartsliderwithrumble) and the chip's data sheet, with a small amount of RE of the game.


windwakr:
OK, I ended up rewriting the slot2 code to make it not so awful, taking heavy inspiration from the [Magic Reader serial handling code](https://github.com/shonumi/gbe-plus/blob/7986317958a672d72ff54ea33f31bcca13cf8330/src/nds/slot2.cpp#L155) in GBE+.

I feel like the slot2 code is in acceptable state now. The frontend code is unchanged and still crap and broken, though.

windwakr:
Ok, last change before leaving this for a while. I added rumble support. In doing this, I moved the definition of FeedbackON from slot2_rumblepak.cpp to slot2.cpp to make it more clear that it's available for other add-ons.

Rumble must be explicitly enabled by the frontend. In my crappy Windows frontend implementation it just checks the ini for a setting. Add this to your ini to enable it.
```
[Slot2.SlideController]
rumble=1
```
Then I guess set your controller on your lap or something? Yeah, it isn't the most practical to use.

PR #764, opened 2024-01-20
gtk: implement GPU scale factor feature

thesourcehim:
WIP implementation of GPU scale factor for linux GTK3 frontend. Currently UI is changed to properly set new option, however scaling itself does not fully work yet: rendered image seems to be scaled but put incorrectly to the window, parts of the image are mixed, like origin point is set somewhere in the middle. Still trying to figure this out, hints are appreciated.


desmume/src/frontend/posix/gtk/graphics.ui
```diff
@@ -94,14 +94,49 @@
...
+              </object>
+              <packing>
+                <property name="left_attach">0</property>
+                <property name="top_attach">2</property>
+              </packing>
+            </child>
+            <child>
+              <object class="GtkComboBoxText" id="gpuscale">
+                <items>
+                  <item translatable="yes">×1</item>
```
rofl0r, line 109: wouldn't it be better to make the scale factor a float, like in cli, and use some sort of editbox with a knob for the selection, instead of hardcoding a few ints in a combobox ?

thesourcehim:
Replaced combobox with spin button and fixed screenshot crash. Did some tests to check if incorrect image comes from GPU or VideoFilter and it looks like it comes from GPU. Maybe, OGL renderer bug? I'll continue digging.

desmume/src/frontend/posix/gtk/main.cpp
```diff
@@ -2087,10 +2097,16 @@ static void GraphicsSettingsDialog(GSimpleAction *action, GVariant *parameter, g
 		default:
 			break;
 		}
-		gpu_scale_factor = gtk_combo_box_get_active(wGPUScale)+1;
+		gpu_scale_factor = gtk_spin_button_get_value(wGPUScale);
+		if(gpu_scale_factor < GPU_SCALE_FACTOR_MIN)
+			gpu_scale_factor = GPU_SCALE_FACTOR_MIN;
+		if(gpu_scale_factor > GPU_SCALE_FACTOR_MAX)
+			gpu_scale_factor = GPU_SCALE_FACTOR_MAX;
```
rofl0r, line 2104: might want to save the normalized value back into the spin button in case it exceeds the limits

thesourcehim:
Yes, I understood what you did in cli, but I'm trying to render at higher resolution using large buffer like windows frontend does. Simply scaling rendered image to window is already supported.

thesourcehim:
Finally fixed incorrect image, turns out, I was using wrong pointer to access the buffer. Also added some sanity checks for scale factor value. For now I changed graphics dialog so only integer values are allowed: testing revealed that non-integer values tend to cause random crashes.

Here's the screenshot rendered at 5x:
desmume-screenshot-0

thesourcehim:
Small fix for OSD when scale factor is > 1. Makes OSD elements at least visible, but unfortunately not scaled. As far as I can see, reworking of AGG OSD code is required since native screen size is hardcoded there in a lot of places. Probably should be done in different PR.

thesourcehim:
Ok, for now I think it's finished. If anyone has other suggestions, I'll be glad to hear them.

rofl0r:
cool. are you prepared to do some follow-up work in case a bug in this code gets detected ? i haven't seen the maintainer of the gtk+3 port for quite some time.
oh, and please provide a screenshot of where you placed the button in the gui, so i can try to port this over to gtk+2.

thesourcehim:
> cool. are you prepared to do some follow-up work in case a bug in this code gets detected ? i haven't seen the maintainer of the gtk+3 port for quite some time. 

Sure. I also have a couple more ideas for GTK3 frontend for the future. AGG OSD needs some touch and I want to implement hybrid layout (similar to melonDS, but with different size preferences).

> oh, and please provide a screenshot of where you placed the button in the gui, so i can try to port this over to gtk+2.

Screenshot_20240125_093908

PR #769, opened 2024-02-01
GTK: make OSD scalable

thesourcehim:
My previous PR (implement GPU scale factor for GTK) created a problem with OSD: it is drawn without scaling, so when scale factor is > 1, OSD becomes small. I implemented OSD scaling. Scaling most elements like rectangles and ellipses seems trivial, but I could not figure out how to scale raster fonts. So instead I used vector fonts. Most of the code to work with vector fonts was already in AGG OSD files, I only added some missing functions and made using vector fonts optional via flag. By default vector fonts are disabled to keep other frontends working as before. It is frontend's job to provide a font file and set this flag. In GTK frontend I use fontconfig to locate monospace font file in the system. Also for vector fonts I could not figure out how to set background, so I made outlines instead.

Things to do:
Properly save OSD layout upon editing, restore on start.

Screenshot with OSD rendered at 4x:
scaled_osd_screenshot

### DIFF
--- a/desmume/src/frontend/modules/osd/agg/agg_osd.cpp
+++ b/desmume/src/frontend/modules/osd/agg/agg_osd.cpp
@@ -561,24 +561,24 @@ static void DrawStateSlots(){
 
 	if(alpha!=0)
 	{
-		aggDraw.hud->lineWidth(1.0);
+		aggDraw.hud->lineWidth(osd->scale);
 		aggDraw.hud->lineColor(0, 0, 0, alpha);
 		aggDraw.hud->fillColor(255, 255, 255, alpha);
 
-		for ( int i = 0, xpos=0; i < 10; xpos=xpos+24) {
+		for ( int i = 0, xpos=0; i < 10; xpos=xpos+24*osd->scale) {
 
 			int yheight=0;
 
-			aggDraw.hud->fillLinearGradient(xloc + xpos, yloc - yheight, xloc + 22 + xpos, yloc + 20 + yheight+20, agg::rgba8(100,200,255,alpha), agg::rgba8(255,255,255,0));
+			aggDraw.hud->fillLinearGradient(xloc + xpos, yloc - yheight, xloc + 22*osd->scale + xpos, yloc + 20*osd->scale + yheight+20*osd->scale, agg::rgba8(100,200,255,alpha), agg::rgba8(255,255,255,0));
 
 			if(lastSaveState == i) {
-				yheight = 5;
-				aggDraw.hud->fillLinearGradient(xloc + xpos, yloc - yheight, 22 + xloc + xpos, yloc + 20 + yheight+20, agg::rgba8(100,255,255,alpha), agg::rgba8(255,255,255,0));
+				yheight = 5*osd->scale;
+				aggDraw.hud->fillLinearGradient(xloc + xpos, yloc - yheight, 22*osd->scale + xloc + xpos, yloc + 20*osd->scale + yheight+20*osd->scale, agg::rgba8(100,255,255,alpha), agg::rgba8(255,255,255,0));
 			}
 
-			aggDraw.hud->rectangle(xloc + xpos , yloc - yheight, xloc + 22 + xpos , yloc + 20 + yheight);
+			aggDraw.hud->rectangle(xloc + xpos , yloc - yheight, xloc + 22*osd->scale + xpos , yloc + 20*osd->scale + yheight);
 			snprintf(number, 10, "%d", i);
-			RenderTextAutoVector(xloc + 1 + xpos + 4, yloc+4, std::string(number), true, osd ? osd->scale : 1);
+			RenderTextAutoVector(xloc + osd->scale + xpos + 4*osd->scale, yloc+4*osd->scale, std::string(number), true, osd->scale);
 			i++;
 		}
 	}
--- a/desmume/src/frontend/modules/osd/agg/agg_osd.cpp
+++ b/desmume/src/frontend/modules/osd/agg/agg_osd.cpp
@@ -89,7 +89,7 @@ static T calcY(T y) // alters a GUI element y coordinate as necessary to obey sw
 	return y;
 }
 
-static void RenderTextAutoVector(double x, double y, const std::string& str, bool shadow = true, int shadowOffset = 1)
+static void RenderTextAutoVector(double x, double y, const std::string& str, bool shadow = true, double shadowOffset = 1.0)
 {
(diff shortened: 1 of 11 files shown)


desmume/src/frontend/modules/osd/agg/agg_osd.cpp
```diff
@@ -624,9 +709,11 @@ OSDCLASS::OSDCLASS(u8 core)
 
 	singleScreen = false;
 	swapScreens = false;
-
+	scale=1;
```
rofl0r, line 712: good idea to make it an int here, when it's a float elsewhere ? (even if currently fractions cause graphics glitches, but that might be fixed at some point)

rofl0r:
nice work, though the amount of changes is rather excessive, wouldn't it be possible (and much easier, even preserving the ability to use pixel fonts) to just make the OSD code render to a transparent SDL_Texture or equivalent at 1x, and then just make SDL2 blit it using its own scaler over the renderer texture ?

desmume/src/frontend/modules/osd/agg/agg_osd.cpp
thesourcehim: Good point, I'll change to double.

thesourcehim:
> nice work, though the amount of changes is rather excessive, wouldn't it be possible (and much easier, even preserving the ability to use pixel fonts) to just make the OSD code render to a transparent SDL_Texture or equivalent at 1x, and then just make SDL2 blit it using its own scaler over the renderer texture ?

I can try that, but I suspect the result will be the same as just settings scale to 1 and scaling entire image to window: HUD becomes blurry, while this scaling draws nice pad and other elements. But maybe I should use raster fonts when not scaling - they look nicer on low resolution.

thesourcehim:
I added saving and loading layout. Also played around with fonts some more: definitely prefer raster font on native resolution - looks much nicer, and choose vector font size close to raster one for consistency.

Screenshot with latest modifications (4x rendering)
Screenshot_20240202_185337

thesourcehim:
I think I finished this one, but I recommend someone else to test it under linux to make sure fonts load properly, though I used two  very common ones, so there should not be any problem.

desmume/src/frontend/modules/osd/agg/aggdraw.cpp
```diff
@@ -162,6 +162,23 @@ void AggDraw_Desmume::setTarget(AggTarget newTarget)
...
+void Agg_setCustomSize(int w, int h)
+{
+	hudBuffer.resize(w*h);
+	luaBuffer.resize(w*h);
+	if(aggDraw.screen)
+	{
+#if defined(WIN32) || defined(HOST_LINUX)
+		aggDraw.screen->setDrawTargetDims(0, w, h, w*4);
+#else
+		aggDraw.screen->setDrawTargetDims(0, w, h, w*2);
```
rofl0r, line 174: what's the difference between linux/win and everything else here ?

desmume/src/frontend/posix/meson.build
```diff
@@ -10,6 +10,8 @@ project('desmume',
   license: 'GPL2+',
 )
 
+add_global_arguments('-DHOST_LINUX', language: ['c', 'cpp'])
```
rofl0r, line 13: is it wise to set this for everyone using meson / gtk3 version ?

desmume/src/frontend/posix/meson.build
```diff
@@ -203,6 +205,10 @@ endif
 if dep_agg.found()
   dependencies += dep_agg
   add_global_arguments('-DHAVE_LIBAGG', language: ['c', 'cpp'])
+  if get_option('frontend-gtk')
+    add_global_arguments('-DAGG2D_USE_VECTORFONTS', language: ['c', 'cpp'])
```
rofl0r, line 209: you add this unconditionally, so i wonder if it still compiles if not set. also how does it look like if not set ?

desmume/src/frontend/modules/osd/agg/aggdraw.cpp
thesourcehim: See also line 121, was not added by me, I only mimiced this in Agg_setCustomSize. If neither WIN32 nor HOST_LINUX is defined than screen buffer is for some reason assumed to be 16bit, which causes HUD to be drawn incorrectly. I don't know what original author ment by this, for which device. May be it's better to use define to specify buffer bitness instead of checking for some platforms?

desmume/src/frontend/posix/meson.build
thesourcehim: Maybe it's better to add meson option to use vector fonts then?

rofl0r: i'd enable it unconditionally if fontconfig lib is found

desmume/src/frontend/modules/osd/agg/aggdraw.cpp
rofl0r: sounds reasonable. we can store byteperpixel used to initialize SDL2, and use it for math rather than hardcoding numbers nobody knows what they mean.

thesourcehim: Added bytes per pixel detection, defaults to 4 (at start and in case SDL call fails for some reason).

desmume/src/frontend/posix/meson.build
thesourcehim: > i'd enable it unconditionally if fontconfig lib is found

Done

desmume/src/frontend/modules/osd/agg/aggdraw.cpp
```diff
@@ -118,11 +118,7 @@ static void Agg_init_fonts()
 
 AggDraw_Desmume aggDraw;
 
-#if defined(WIN32) || defined(HOST_LINUX)
 T_AGG_RGBA agg_targetScreen(0, GPU_FRAMEBUFFER_NATIVE_WIDTH, GPU_FRAMEBUFFER_NATIVE_HEIGHT*2, 1024);
-#else
```
rofl0r, line 123: shouldn't this now read `256*aggDraw.screenBytesPerPixel` instead of 1024 ?

oh, actually if bpp == 2, it should return T_AGG_RGB555 instead of T_AGG_RGBA

thesourcehim: You're right. The problem is it's not a function, it's global object. Made a cheap workaround. Ideally this object should be allocated dynamically, but it can severely break other frontends.

PR #773, opened 2024-02-21
GTK: hybrid layout support

thesourcehim:
Introducing two additional layout types: "Hybrid (equal height)" and "Hybrid (emphasize vertical)".
Both are basically concatenation of single screen layout and vertical layout.
The first one behaves like in melonDS (I think): single (big) screen and vertical part have equal heights and resulting concatenated rectangle is fit into drawing area.

hybrid_equal

The second one is a bit trickier: vertical part is scaled to occupy all available vertical space (if possible) and placed on the right, then big screen is fit into remaining part. Big screen is aligned at the top of the area if it's the top screen and at the bottom if it's the touch screen.

hybrid_vertical_top

hybrid_vertical_bottom

This type of layout allows vertical part to be as large as possible. It's specifically designed for wide screens.

In both cases 'Swap screens' option controls wich screen is displayed as big screen, it does not affect vertical part. Other options like gap and rotation are respected.

Additionally I optimized drawing function a bit, so recalculation of some parameters only occurs when necessary, not every frame.

### DIFF
--- a/desmume/src/frontend/posix/gtk/main.cpp
+++ b/desmume/src/frontend/posix/gtk/main.cpp
@@ -313,6 +313,8 @@ enum orientation_enum {
     ORIENT_VERTICAL = 0,
     ORIENT_HORIZONTAL = 1,
     ORIENT_SINGLE = 2,
+    ORIENT_HYBRID_EQUAL = 3,
+    ORIENT_HYBRID_VERTICAL = 4,
     ORIENT_N
 };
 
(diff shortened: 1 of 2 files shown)


thesourcehim:
Requesting review. @rofl0r ?

desmume/src/frontend/posix/gtk/main.cpp
```diff
@@ -3464,6 +3662,12 @@ common_gtk_main(GApplication *app, gpointer user_data)
         case ORIENT_SINGLE:
             string = "single";
             break;
+        case ORIENT_HYBRID_EQUAL:
+			string = "hybrid_equal";
+			break;
+		case ORIENT_HYBRID_VERTICAL:
```
rofl0r, line 3668: tabs vs spaces

thesourcehim: Oops, yeah, sorry, fixed.

rofl0r:
cool. i assume you tested this sufficiently ? i deliberately waited a few days after you first opened this, and indeed your subsequent commits proved me right.

thesourcehim:
I tested this with various combinations of layout, rotation, screen gap and swap screens options and with different window sizes (including different horizontal/vertical proportions of the window). After that I tested HUD including edit mode and additional commits were required. For instance, with hybrid layouts 'Swap screens' flag should not be transfered to HUD because it does not affect vertical part and if transfered, HUD parts would jump between screens. Now I tested HUD with all layout types and it looks like display and edit work properly now. If you have other tests in mind I will perform them.

PR #784, opened 2024-04-09
add an issue template for github

rofl0r:
since 90% of opened issues are totally low quality, not even stating their OS, hopefully this template improves things.


rofl0r:
@zeromus fine with adding this ?

zeromus:
Yeah sure it's fine, but IMO it's better to ask people exactly which desmume URL they downloaded, because if you ask them "which version" they will say "latest" or something and be wrong.

rofl0r:
alright, then let's see how this goes. adding a question for download url is a matter of adding 2 lines.

PR #792, opened 2024-05-13
libretro-common: fix implicit declarations

atsampson:
strdup and realpath are only declared by glibc's headers if _XOPEN_SOURCE >= 500.

(I checked libretro-common upstream, and this doesn't seem to have been fixed there; maybe they're not building with -std=c99 or similar. Previously it would just have produced a warning but an implicit declaration is a hard error with GCC 14.)

### DIFF
--- a/desmume/src/libretro-common/file/file_path.c
+++ b/desmume/src/libretro-common/file/file_path.c
@@ -20,6 +20,8 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
+#define _XOPEN_SOURCE 500   /* For strdup, realpath */
+
 #include <stdlib.h>
 #include <boolean.h>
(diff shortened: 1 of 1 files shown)


zeromus:
that broke the mac build. strlcpy this time. increase the version number or something.

rofl0r:
usually 700 is the lowest value one uses for the XOPEN feature test macro.

atsampson:
Looking at macOS's [cdefs.h](https://opensource.apple.com/source/xnu/xnu-4570.31.3/bsd/sys/cdefs.h.auto.html) and [string.h](https://opensource.apple.com/source/Libc/Libc-1439.141.1/include/string.h.auto.html), it looks like `strlcpy` is only defined if `__DARWIN_C_LEVEL >= 900000`, so there's no value of `_XOPEN_SOURCE` that would enable that definition. It might need an `#ifdef` for macOS to define `_DARWIN_C_SOURCE` there...

Issue #812, opened 2024-07-15
GLES: GL_OES_surfaceless_context missing


thesourcehim:
## State your operating system:
Linux/gtk3

## DesMuME version
master

## Issue
I'm trying to add GLES support to gtk frontend, however GLES renderer keeps throwing error about missing extension GL_OES_surfaceless_context. This extension is really not present for both my RTX 3060 and Intel integrated graphics despite having latest drivers. I created a PR, maybe I'm simply doing something wrong.

Is this extension really required to use FBO's?

rogerman:
Yes, GL_OES_surfaceless_context will be required at the moment, until I can prove that it isn't required. But to prove it, I'll need to fix some of the fundamental problems with the OpenGL ES renderer first.

After I'm happy about the ES renderer's compatibility, I'll look into removing the requirement for GL_OES_surfaceless_context.

PR #813, opened 2024-07-15
[WiP] GTK: add OpenGL ES support

thesourcehim:
Added support for GLES in posix/gtk frontend. Does not work for now, initialization fails with missing extension. See #812.

### DIFF
--- a/desmume/src/frontend/posix/gtk/graphics.ui
+++ b/desmume/src/frontend/posix/gtk/graphics.ui
@@ -65,6 +65,7 @@
                   <item translatable="yes">Null</item>
                   <item translatable="yes">SoftRasterizer</item>
                   <item translatable="yes">OpenGL</item>
+				  <item translatable="yes">OpenGL ES</item>
                 </items>
               </object>
               <packing>
--- a/desmume/src/frontend/posix/gtk/graphics.ui
+++ b/desmume/src/frontend/posix/gtk/graphics.ui
@@ -65,7 +65,7 @@
                   <item translatable="yes">Null</item>
                   <item translatable="yes">SoftRasterizer</item>
                   <item translatable="yes">OpenGL</item>
-				  <item translatable="yes">OpenGL ES</item>
+                  <item translatable="yes">OpenGL ES</item>
                 </items>
               </object>
               <packing>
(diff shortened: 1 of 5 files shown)


thesourcehim:
Figured what was wrong with initialization. Starts properly now, but the image renders incorrectly, like it is partially visible behind black surface

gles_test

desmume/src/frontend/posix/gtk/graphics.ui
```diff
@@ -65,6 +65,7 @@
                   <item translatable="yes">Null</item>
                   <item translatable="yes">SoftRasterizer</item>
                   <item translatable="yes">OpenGL</item>
+				  <item translatable="yes">OpenGL ES</item>
```
rofl0r, line 68: you seem to have mixed tabs and spaces here

thesourcehim: Fixed

thesourcehim:
Setting multisampling to something other than 'None' makes problem less apparent. Also 2d renders fine it looks like, it's only 3d elements that hide behind blackness.

rogerman:
Sorry, but there are some fundamental problems with the OpenGL ES renderer at the moment, which is why no port was updated to use it just yet. But I'm getting there. Just wait a few days and check back.

thesourcehim:
I just tested with commit 60385bd and the situation has improved somewhat: less black tint on nvidia gpu, but still present
nvidia
And it seems that the higher the fps the more black tint is on screen (threading issue?)

On intel gpu some blue dots appeared over blackness, but overall everything is still black on intel.

rogerman:
Thanks for checking, but the ES renderer isn't meant to be working 100% just yet. It's still in a testing phase. As of now, there is still one major fundamental problem that exists.

When you see the "egl_3Demu.cpp" and "egl_3Demu.h" files appear under gtk2, then you'll know that the OpenGL ES renderer is "done".

rogerman:
The ES renderer is now complete as of commit be51e41. I've updated all of the POSIX frontends to work with the new ES renderer using some of your work as inspiration.

- If you're building with Autotools, you'll need to use --enable-opengles
- If you're building with Meson, you'll need to set -Dopengles=true

PR #822, opened 2024-08-08
CLI port: Allow to map NDS keys to joystick buttons.

MasterEnoc:
- It allows to map NDS key to joystick buttons.  Keys were mapped and
then restored with load_default_config(), thus I remove it, however,
looks like gtk frontends have a different keycode for joystick
buttons, thus a user of both frontends will have to rewrite the
config each time it uses one frontend or the other.

- it fixes config file processing, it now continues even when
errors are found. Previously, mapping a NDS button to a joysticks
buttons had to be sequencially, I. e. A, then B, then Select, etc. If
any key was skipped in the corresponding order desmume would stop from
mapping the keys.

- Add boost support for joysticks, currently it works in a hold
  fashion, I do not know if it should be changed to toggle, or
  both. Additionally, I implemented this in a very cheap way, I did not want 
  to make big chances in the codebase, nonetheless, I could try to fix it in other
  way if you do not like it.


desmume/src/frontend/posix/shared/ctrlssdl.cpp
```diff
@@ -430,20 +432,23 @@ do_process_joystick_events( u16 *keypad, SDL_Event *event) {
...
-        ADD_KEY( *keypad, key );
+          ADD_KEY( *keypad, key );
       break;
 
-      /* Joystick button released */
+        /* Joystick button released */
     case SDL_JOYBUTTONUP:
       key_code = ((event->jbutton.which & 15) << 12) | JOY_BUTTON << 8 | (event->jbutton.button & 255);
       key = lookup_joy_key( key_code );
+      if (key == 0x2000)
```
rofl0r, line 448: what's that magic constant ?

desmume/src/frontend/posix/cli/main.cpp
```diff
@@ -314,15 +314,15 @@ static void desmume_cycle(struct ctrls_event_config * cfg)
         (SDL_PollEvent(&event) || (!cfg->focused && SDL_WaitEvent(&event))))
       {
         process_ctrls_event( event, cfg);
-    }
+      }
```
rofl0r, line 317: please dont gratuitously change whitespace, even if it appears wrong this should be done as a separate cleanup PR.

desmume/src/frontend/posix/cli/main.cpp
```diff
@@ -314,15 +314,15 @@ static void desmume_cycle(struct ctrls_event_config * cfg)
...
 
     /* Update mouse position and click */
     if(mouse.down) {
 	NDS_setTouchPos(mouse.x, mouse.y);
 	mouse.down = 2;
     }
     if(mouse.click)
-      { 
+      {
```
rofl0r, line 325: likewise

desmume/src/frontend/posix/cli/main.cpp
```diff
@@ -427,7 +427,7 @@ int main(int argc, char ** argv) {
 			slot2_device_type = NDS_SLOT2_NONE;
 			break;
 	}
-    
+
```
rofl0r, line 430: likewise, etc

MasterEnoc:
I fixed whitespace issues, and added a comment about the magic number in ctrlssdl.cpp.

MasterEnoc:
I missed one whitespace, all should be fixed now.

zeromus:
Fixing a pointless whitespace commit with another commit defeats the purpose, unless the PR is squashed. The right way to do this would be to rebuild the patch as a single commit. It's a good exercise for the future if you do that with git tools (soft reset to master on the commit prior to your work, then commit the current differences as a new commit) because it's helpful for cleaning up more severe changes, but you get the same result if you clone the repo again and paste your working copy on top and create a new commit. In either case once done you would have to force push it to the PR branch.

MasterEnoc:
Fixed, just rebased the branch on the current master and squashed the last two commits (the ones about whitespace).

PR #839, opened 2024-09-26
Update org.desmume.DeSmuME.metainfo.xml

intra0:
the current flatpack version is commit 640a1fdd93ac935314ab2a69725d03abc0ba8a8b

was 2 years ago but is a 9.14 version


intra0:
@zeromus can this update just be implemented

Issue #843, opened 2024-10-06
[linux] editing controlls sometimes (often) freezes the emulator


intra0:
Tested using the newest commit as of october 2024, built using gtk3
![Screenshot from 2024-10-05 19-34-37](https://github.com/user-attachments/assets/e18efa76-85e2-4b7d-a0ed-fea0bc645684)
when editing controlls, if you are not quick enough (and 60% of the time even if you are) the emulator will hard freeze.
music will still play, and the game will run, and if you already have some controls set up by the time it freezes, you can use those buttons.
but when I say it hard freezes, I mean it becomes impossible to quit the program. none of the UI is interactable, and alt+f4 does nothing. only thing that works is opening up the task manager and killing the process.

intra0:
This issue was reported for the flatpak version in #731 but it happens on *all* linux versions of desmume

thesourcehim:
I tried to reproduce using my local build, but no luck so far. Is it exclusive to flatpack version?

intra0:
No I compiled the latest commit from source and reproduced the issue
also happens in the fedora rpm fusion version

try opening the change A button and leaving it for like 30 seconds to a minute

thesourcehim:
I started a game, then opened 'Configure joystick controls' window, pressed a button at the top (configure A), the emulation paused, which is expected: the dialog appears waiting for the input in a loop in the main thread. I waited for a minute then pressed a button on my gamepad, input was accepted and emulation continued.

rofl0r:
wayland vs X11 issue maybe ?

intra0:
I am using wayland. what distro are you using @thesourcehim ?

intra0:
and it only happens for joystick controlls, does not happen for keyboard controlls, are you using keyboard @thesourcehim ?

changing the focused window while this is up can also cause the freeze
![image](https://github.com/user-attachments/assets/a63a5956-9835-4bb0-8da0-4ac817af5256)

thesourcehim:
Yes, I'm using gamepad. My system is Fedora 40, Plasma 6 wayland.
You can not press Ok or Cancel until you give some input from your gamepad because as I explained earlier application enters a loop waiting for input. This can lead to a deadlock if for example gamepad gets disconnected while app is in the loop (hotplugging is not supported). I agree this is not the best way to configure controls, I'll rework that dialog to get input asynchronously. For now you have press something on a gamepad to exit the loop.

thesourcehim:
[...]

Issue #834, opened 2024-09-22
[LINUX/GTK3] Whitescreen freeze when pressing square on ds4


Connor-Kolan:
## State your operating system:
Linux CLI and gtk3 

## DesMuME version
0.9.13

## Issue
I play with a ds4 controller and have used steam and AntiMicroX to remap its controls to the keyboard, the bug occurs on both.
I bound the square button to the "J"-Key on my keyboard which I then bound to the "Y" Key in desmume.
After loading a rom and pressing the square button, desmume seems to pause/freeze by switching to a white screeen and playing the same audio frame all the time. I can then press the square button again and it returns to normal. This only happens when I press press square on the controller. If I press the "J"-Key it works just fine.

thesourcehim:
Check if square button is bound to anything in the emulator settings, maybe something else triggers beside Y. Also please attach here your AntiMicroX configuration (amgp file).

Connor-Kolan:
The button is not bound to anything else. What I have noticed is, that the error seems to only occur when using the controller in wired mode, when using bluetooth the bug does not occur. 
I wont let me upload the file with the original file ending so I renamed it to txt but this is the AntiMicroX config
[t.gamecontroller.txt](https://github.com/user-attachments/files/17134684/t.gamecontroller.txt)

Connor-Kolan:
I've tried it again and it seems to only happen when I start desmume for the second time after connecting my controller. If I reconnect my controller it works again.

thesourcehim:
@Connor-Kolan I just came back from a buisness trip, I'll test with DualSense (I don't have DS4) as soon as I can.

Connor-Kolan:
Steps to recreate: 

1. Connect controller
2. start desmume
3. boot rom
4. press square 
5. white screen

Steps to fix:

1. disconnect controller while desmume is running
2. reconnect controller 

My config file: (again it wont let me upload it without renaming it to txt)
[config.txt](https://github.com/user-attachments/files/17292797/config.txt)

thesourcehim:
I'll test tomorrow but I can say that reconnecting controller while desmume is running changes things because gtk frontend does not support gamepad hotplugging. So after you reconnect controller desmume only gets keyboard input from antimicrox, before reconnecting it gets both J and square

thesourcehim:
[...]

PR #845, opened 2024-10-10
[Linux/GTK3] Some gamepad input rework

thesourcehim:
1. Unbind gamepad keys by default. The default bindings may fit one gamepad model but work weirdly with another causing issues like #834
2. Use non-blocking method to obtain gamepad keys/axes during configuration to avoid visible emulator freeze and possible deadlock (see #843)

### DIFF
--- a/desmume/src/frontend/posix/gtk/main.cpp
+++ b/desmume/src/frontend/posix/gtk/main.cpp
@@ -115,6 +115,7 @@ int real_framebuffer_height = GPU_FRAMEBUFFER_NATIVE_HEIGHT;
 
 
 desmume::config::Config config;
+bool in_joy_config_mode = false;
 
 #ifdef GDB_STUB
 gdbstub_handle_t arm9_gdb_stub = NULL;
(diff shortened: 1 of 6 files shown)


intra0:
I can confirm that completely this fixes #843 on my end

rofl0r:
@thesourcehim does this also supersede #822 ? if not, maybe you can take a look at whether the changes there make sense.

thesourcehim:
In #822 I see changes related to how input config file is loaded and joystick events are processed for the main emu loop, some changes to cli frontend. My PR changes the default mappings for gamepad should config file lack gamepad input config, I didn't touch config file load algorithm. Also reworks input config dialog for GTK3 frontend. So I think these two PRs have different purpose.

intra0:
#822 never touches the gtk file, and this commit never touches the cli file
and the only change this pr has to ctrlssdl.cpp is changing the default button values, where #822 makes changes to code in that file and doesn't touch default button values.
@rofl0r these are completely different and unrelated PR's and I dont think there are any changes here that would conflict with #822

rofl0r:
@thesourcehim would you be able/willing to port this to GTK2 ? that way i could test it and merge if everything's fine.

thesourcehim:
> @thesourcehim would you be able/willing to port this to GTK2 ? that way i could test it and merge if everything's fine.

Ok, I can do it.

thesourcehim:
@rofl0r done.

rofl0r:
thanks, works quite well.
imo the joystick config could be further improved by not having to click on OK after every press, but i suppose that's not that straightforward to do with standard GTK controls.

rofl0r:
another thing is i'd recommend to use a feature branch for each PR you make, instead of pushing to your master - it doesn't mess up your git clone after some work was merged with changes, and it makes it easier for maintainers to check out your changes, rebase, etc.

PR #846, opened 2024-10-10
Cocoa Port: Fix mixed retina/non-retina display touchscreen bug

BayLee4:
This pull request fix a bug where the touchscreen coordinates were erroneous when using a mixed retina/non-retina display setup on macOS. I don't know if those defines are still necessary, but I kept them for code consistency.


rogerman:
@BayLee4, good catch! This is a good PR to approve as-is.

And yes, those defines are still necessary, since it is possible to build Mac DeSmuME in Xcode 3 on a PowerPC Mac.

PR #847, opened 2024-10-11
[Linux - GTK3/GTK2] Action Replay code menu (+ other cheat menu improvements)

### LINKED ISSUE #725: Cheat offsets are decimal in Linux Gtk frontend
I tried to enter an offset of `1206E120` and desmume would always cut it off at the occurrence of the first letter, `E`, making it `1206`. I figured out that the field for offset input is actually expecting decimal numbers, rather than hexadecimal.

En-En-Code:
Summary:
The Linux GTK frontend's cheat menu now reads Action Replay codes from a game's `dct` file and adds them to a seperate section where users can add, remove, toggle, and modify them correctly and conveniently.
A couple changes were also made to the internal cheats' section of the menu to make it harder to misuse:
 - The "Offset" field has been changed to an "Address" field, and now expects 7 nibbles of hexadecimal data (fixes #725).
 - The value can no longer be outside of its logical range (for example, a 1-byte cheat is capped at 255).

Notes:
Action Replay code entry still leaves something to be desired. The scroll window sometimes doesn't scroll all of the way down and there is no size cut-offs, so long codes can consume far too much space. Coding two new GTK objects to get multiline editable text was easily the hardest part of this coding adventure though.
I still need to backport much to the GTK2 frontend, but with a working GTK2 build, I expect that part to be merely tedious, not actually hard.
The formatting was fairly scattered throughout the sections I worked on, (noticable really early on, when Neovim picked up the `.editorconfig` and inserted tabs, which I usually set to 2 spaces). I wrote an opinionated `.clang-format` that kept at least some of the formatting decisions I gleaned (but not tabs, screw tabs). I'll share it if you want it.
I wanted to cheat something for a fun project, which led me to realizing the shortcomings of the internal cheat menu. After converting to decimal and discovering they were not capable of doing what I wanted, I found the file cheats were stored, added Action Replay codes to another file on a Windows version, then copied them back to the `dct` file on Linux. Seeing they worked flawlessly meant everything in the backend was there, so it was just interface programming. Found the culprit `atoi` messing with internal cheat offset in like 2 minutes, and though the rest was not as easy, its still really nice at the end to tangibly improve a project because it's open source. <3

### DIFF
--- a/desmume/src/cheatSystem.h
+++ b/desmume/src/cheatSystem.h
@@ -119,6 +119,9 @@ public:
 	bool	update_CB(char *code, char *description, u8 enabled, const size_t pos);
 	
 	bool	remove(const size_t pos);
+
+	void	toggle(bool enabled, const size_t pos);
+	void	toggle(u8 enablbed, const size_t pos);
 	
 	void	getListReset();
 	bool	getList(CHEATS_LIST *cheat);
--- a/desmume/src/cheatSystem.h
+++ b/desmume/src/cheatSystem.h
@@ -1,5 +1,5 @@
 /*
-	Copyright (C) 2009-2023 DeSmuME team
+	Copyright (C) 2009-2024 DeSmuME team
 
 	This file is free software: you can redistribute it and/or modify
 	it under the terms of the GNU General Public License as published by
@@ -119,7 +119,7 @@ public:
 	bool	update_CB(char *code, char *description, u8 enabled, const size_t pos);
 	
 	bool	remove(const size_t pos);
-
+	
 	void	toggle(bool enabled, const size_t pos);
 	void	toggle(u8 enablbed, const size_t pos);
 	
--- a/desmume/src/frontend/posix/gtk/cheatsGTK.cpp
+++ b/desmume/src/frontend/posix/gtk/cheatsGTK.cpp
@@ -58,7 +58,7 @@ static struct {
 } columnTable[]={
     { "Enabled", TYPE_TOGGLE, COLUMN_ENABLED},
     { "Size", TYPE_COMBO, COLUMN_SIZE},
-    { "Offset", TYPE_STRING, COLUMN_HI},
+    { "Address", TYPE_STRING, COLUMN_HI},
     { "Value", TYPE_STRING, COLUMN_LO},
     { "Description", TYPE_STRING, COLUMN_DESC}
 };
@@ -91,7 +91,7 @@ enabled_toggled(GtkCellRendererToggle * cell,
     path1 = gtk_tree_model_get_path (model, &iter);
     ii = gtk_tree_path_get_indices (path)[0];
 
-	cheats->copyItemFromIndex(ii, tempCheatItem);
+    cheats->copyItemFromIndex(ii, tempCheatItem);
 
     cheats->update(tempCheatItem.size, tempCheatItem.code[0][0], tempCheatItem.code[0][1], tempCheatItem.description,
                  cheatEnabled, ii);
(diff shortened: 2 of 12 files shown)


desmume/src/cheatSystem.h
```diff
@@ -120,6 +120,9 @@ class CHEATS
 	
 	bool	remove(const size_t pos);
 	
+	void	toggle(bool enabled, const size_t pos);
+	void	toggle(u8 enablbed, const size_t pos);
```
rofl0r, line 124: typo. do we really need an overloaded func for this ?

desmume/src/frontend/posix/gtk/cheatsGTK.cpp
```diff
@@ -18,178 +18,237 @@
...
+    COLUMN_AR,
     COLUMN_DESC,
     NUM_COL
 };
 
 enum
 {
-  COLUMN_SIZE_TEXT,
-  NUM_SIZE_COLUMNS
+    COLUMN_SIZE_TEXT,
```
rofl0r, line 48: i'd really appreciate if whitespace changes could be kept to the bare minimum. if everyone reformats the code he's working on to his personal preferences, 90% of every commit is just visual noise when looking at the diffs.

rofl0r:
this is marked as a draft. does this mean that the GTK2 version isn't completed yet ?
i also only see the new TU being used by the GTK3 code.

please also provide some examples of cheat code/game name so this can be tested.

desmume/src/cheatSystem.h
En-En-Code: Maybe not, but the entire cheatSystem overloaded `enabled` this way: `add` [L99-100](https://github.com/TASEmulators/desmume/blob/058fc066c521e1e5f6218a8fcbc211fc37204a74/desmume/src/cheatSystem.h#L99-L100), `update` [L103-104](https://github.com/TASEmulators/desmume/blob/058fc066c521e1e5f6218a8fcbc211fc37204a74/desmume/src/cheatSystem.h#103-L104), `add_AR` [L109-110](https://github.com/TASEmulators/desmume/blob/058fc066c521e1e5f6218a8fcbc211fc37204a74/desmume/src/cheatSystem.h#L109-L110), etc. I followed suit, assuming there was a design reason behind it.

desmume/src/frontend/posix/gtk/cheatsGTK.cpp
En-En-Code: No problem, I'll do that first thing. I just try to leave code less of a mess than when I started.

En-En-Code:
> this is marked as a draft. does this mean that the GTK2 version isn't completed yet ? i also only see the new TU being used by the GTK3 code.

Yes, I realized part way through that repeatedly backporting everything before every commit would be a waste when I commit changes on top of one another, so I decided I could save backporting for the end and make a draft PR with most of the progress.

> please also provide some examples of cheat code/game name so this can be tested.

I like my test suites big, so I will claim anything on [DeSmuMe's wiki page on cheats](https://wiki.desmume.org/index.php?title=Using_Cheats_in_DeSmuMe) will work.

desmume/src/cheatSystem.h
rofl0r: then it's probably fine.

rofl0r:
cool. give me a heads-up when it's ready so i can test the gtk2 version.

En-En-Code:
Alright, GTK2 should be on par with GTK3 functionality. I am still disappointed with the UI with the cheat cell, though it is usable for cheats of reasonable length. There should be a way to bound `CellRenderer`'s maximum height and use a scroll bar for display, but for starting with no GTK knowledge, I cannot complain. It would be an improvement of marginal enough value that I think it should not delay the feature merge.

rofl0r:
finally got around to test the PR, it's a huge improvement - thanks. action replay codes finally work and the input is hex as expected.

however i had to make the following changes:
```diff
--- a/desmume/src/frontend/posix/gtk/utilsGTK.cpp
+++ b/desmume/src/frontend/posix/gtk/utilsGTK.cpp
@@ -19,6 +19,7 @@
  */
 
 #include "utilsGTK.h"
+#include <string.h>
 
 /*
     A C++ implementation of a GtkCellRendererText subclass which handles
diff --git a/desmume/src/frontend/posix/gtk2/utilsGTK.cpp b/desmume/src/frontend/posix/gtk2/utilsGTK.cpp
index 831340e1..f7f893c5 100644
--- a/desmume/src/frontend/posix/gtk2/utilsGTK.cpp
+++ b/desmume/src/frontend/posix/gtk2/utilsGTK.cpp
@@ -20,7 +20,7 @@
 
 #include "utilsGTK.h"
 #include <gdk/gdkkeysyms.h>
-
+#include <string.h>
 /*
     A C++ implementation of a GtkCellRendererText subclass which handles
     newline-delimited text and allows for editing that text with the ability
@@ -56,11 +56,13 @@ static void desmume_entry_nd_cell_editable_init(GtkCellEditableIface *iface);
 
 // As defined in GObject 2.38, which is past the last release of GTK2.
 // https://gitlab.gnome.org/GNOME/glib/-/blob/main/gobject/gtype.h#L2188
+#ifndef G_ADD_PRIVATE
 #define G_ADD_PRIVATE(TypeName)                                                \
     {                                                                          \
         TypeName##_private_offset = g_type_add_instance_private(               \
             g_define_type_id, sizeof(TypeName##Private));                      \
     }
+#endif
 
 G_DEFINE_TYPE_WITH_CODE(DesmumeEntryNd, desmume_entry_nd, GTK_TYPE_EVENT_BOX,
                         G_ADD_PRIVATE(DesmumeEntryNd) G_IMPLEMENT_INTERFACE(
```
this is because you used strlen but didn't include string.h. also the G_blah macro is already defined - glib and gtk2 are independent and it's possible to use a newer glib with gtk2.
please add that patches changes to your fork, then it can be merged.

PR #851, opened 2024-10-20
Fix ensata sound register emulation.

### LINKED ISSUE #850: No Audio when Ensata Emulation Enabled
## State your operating system:
Windows11

## DesMuME version
nightly

## Issue
Games output no audio when ensata emulation is enabled.
Once ensata is disabled, audio works again.

I am not sure if links are allowed here, as they are related to the Nintendo leak occurred in 2020 and GameFreak leak happening recently.

dlbuhtig4096:
Fix #850.

### DIFF
--- a/desmume/src/MMU.cpp
+++ b/desmume/src/MMU.cpp
@@ -2475,6 +2475,7 @@ bool validateIORegsWrite(u32 addr, u8 size, u32 val)
 			if(addrMasked == eng_3D_CLIPMTX_RESULT) return true;
 			if(addrMasked == 0x04FFF000) return true;
 			if(addrMasked == 0x04FFF010) return true;
+			if(addrMasked == 0x04FFF200) return true;
 		}
 
 		switch (addrMasked)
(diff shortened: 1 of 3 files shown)


dlbuhtig4096:
This PR should also fix the problem that soft reset is not working when Ensata emulation enabled now.

Please refers to [comments on the source file](https://github.com/dlbuhtig4096/desmume/blob/master/desmume/src/MMU.cpp#L1846) for the details of the fix.

dlbuhtig4096:
Changed the behavior of OpenConsole() on Windows:
 - First, it attempts to `AttachConsole(-1)` and then uses `AllocConsole()` in case the first attempt fails.
 - Removed `SetConsoleCP(GetACP())` and `SetConsoleOutputCP(GetACP())` upon success.

Most first-party games were developed in Japan, where the log is encoded in CP932 (Shift JIS).
For users not using Japanese Windows, setting the output encoding with GetACP() results in corrupted output.
This change will allow users to set a custom code page for the output console beforehand.

PR #855, opened 2024-10-29
Linux/GTK3: add gamepad hotplug support

thesourcehim:
Support for reconnecting gamepads and connecting new ones while desmume is already running. For that I added corresponding event processing to ctrlssdl, also modifying how connected joystick info is stored (and fixing joystick uninit in the process: previously open_joysticks array was allocated but never populated, but used in uninit function nevertheless), and how pressed button id is calculated (previously joystick instance id was used but it changes every time gamepad is reconnected, so now joystick index is used instead).
This can be backported to GTK2 but GTK2 is now ahead of my branch after recent commits, so please review this PR as it is for now.

### DIFF
--- a/desmume/src/frontend/posix/gtk/main.cpp
+++ b/desmume/src/frontend/posix/gtk/main.cpp
@@ -2088,13 +2088,13 @@ static gboolean JoyKeyAcceptTimerFunc(gpointer data)
         switch(event.type)
         {
         case SDL_JOYBUTTONDOWN:
-            key = ((event.jbutton.which & 15) << 12) | JOY_BUTTON << 8 | (event.jbutton.button & 255);
+            key = ((get_joystick_number_by_id(event.jbutton.which) & 15) << 12) | JOY_BUTTON << 8 | (event.jbutton.button & 255);
             done = TRUE;
             break;
         case SDL_JOYAXISMOTION:
             if( ((u32)abs(event.jaxis.value) >> 14) != 0 )
             {
-                key = ((event.jaxis.which & 15) << 12) | JOY_AXIS << 8 | ((event.jaxis.axis & 127) << 1);
+                key = ((get_joystick_number_by_id(event.jaxis.which) & 15) << 12) | JOY_AXIS << 8 | ((event.jaxis.axis & 127) << 1);
                 if (event.jaxis.value > 0)
                     key |= 1;
                 done = TRUE;
@@ -2102,7 +2102,7 @@ static gboolean JoyKeyAcceptTimerFunc(gpointer data)
             break;
         case SDL_JOYHATMOTION:
             if (event.jhat.value != SDL_HAT_CENTERED) {
-                key = ((event.jhat.which & 15) << 12) | JOY_HAT << 8 | ((event.jhat.hat & 63) << 2);
+                key = ((get_joystick_number_by_id(event.jhat.which) & 15) << 12) | JOY_HAT << 8 | ((event.jhat.hat & 63) << 2);
                 if ((event.jhat.value & SDL_HAT_UP) != 0)
                     key |= JOY_HAT_UP;
                 else if ((event.jhat.value & SDL_HAT_RIGHT) != 0)
@@ -2113,6 +2113,9 @@ static gboolean JoyKeyAcceptTimerFunc(gpointer data)
                     key |= JOY_HAT_LEFT;
                 done = TRUE;
             }
+            break;
+        default:
+            do_process_joystick_device_events(&event);
             break;
         }
     }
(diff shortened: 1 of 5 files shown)


thesourcehim:
@rofl0r If it's ok, I can port this to GTK2 too, so you can test.

desmume/src/frontend/posix/shared/ctrlssdl.cpp
```diff
@@ -39,7 +41,8 @@ u16 nbr_joy;
 mouse_status mouse;
 static int fullscreen;
 
-static SDL_Joystick **open_joysticks = NULL;
+//List of currently connected joysticks (key - instance id, value - SDL_Joystick structure, joystick number)
+static std::unordered_map<SDL_JoystickID, std::pair<SDL_Joystick *, SDL_JoystickID> > open_joysticks;
```
rofl0r, line 45: seems like overkill, currently we support a max of 16 joysticks, which means the joystick id we use fits into one byte... and the average gamer certainly has max 4 joysticks connected so everything would fit into one cache line using a dumb loop over an array. i'm pretty certain the overhead involved in the use of a map for such tiny amount of data is way slower

desmume/src/frontend/posix/gtk/main.cpp
```diff
@@ -4118,6 +4128,7 @@ common_gtk_main(GApplication *app, gpointer user_data)
     video->SetFilterParameteri(VF_PARAM_SCANLINE_D, _scanline_filter_d);
 
 	RedrawScreen();
+	g_timeout_add(200, OutOfLoopJoyDeviceCheckTimerFunc, 0);
```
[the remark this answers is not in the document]
rofl0r: on second thought, why poll at all ? sdl sends an event when a device is plugged in, so we could simply notify the frontend about that, or am i missing something ?

desmume/src/frontend/posix/shared/ctrlssdl.cpp
rofl0r: > which means the joystick id we use fits into one byte

actually, this could be derived from the array index, so it's for free. we only need to make sure to fix the ordering of the array on insertion/removal of a device

desmume/src/frontend/posix/gtk/main.cpp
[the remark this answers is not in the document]
thesourcehim: We must read the event from the queue as far as I understand, but when emulation is not started, events are not processed in the loop (EmuLoop function). When EmuLoop starts, this timer fucntion exits.

desmume/src/frontend/posix/shared/ctrlssdl.cpp
thesourcehim: Then I probably just use static array of 16 elements?

rofl0r: yeah, sounds like the best option. in the unlikely case that more than 16 joys are connected, we just have to discard those > 15.

desmume/src/frontend/posix/shared/ctrlssdl.cpp
```diff
@@ -145,17 +148,13 @@ BOOL init_joy( void) {
...
 {
-  int i;
-
-  if ( open_joysticks != NULL) {
-    for (i = 0; i < SDL_NumJoysticks(); i++) {
-      SDL_JoystickClose( open_joysticks[i]);
-    }
-
-    free( open_joysticks);
+  for (auto & it: open_joysticks) {
```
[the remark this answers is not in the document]
thesourcehim: open_joysticks is now static array and is iterated with int index.

desmume/src/frontend/posix/shared/ctrlssdl.cpp
rofl0r: nbr_joy is used by the "interface" port. if you can add a tiny function that counts the joysticks in the array, the interface can call that and we can get read of the variable

thesourcehim: Added int get_number_of_joysticks()

thesourcehim: Should I also make corresponding changes to gtk2 frontend in my branch?

rofl0r: yes, please

thesourcehim: Done

rofl0r:
thanks, i currently don't have time to test this exhaustively so i'll just go ahead and merge as-is.
if my testing reveals any issues i'm gonna fix it at that point.

Issue #864, opened 2024-11-07
Linux/EGL: prime render offload does not work, iGPU is always used


thesourcehim:
## State your operating system:
Linux Fedora 41, GTK3 frontend, OpenGL 3.2/GLES 3 with EGL

## DesMuME version
git master

## Issue
For some reason when EGL is used to initialize OpenGL or GLES, Intel iGPU is always used despite environment variables are set to use NVIDIA GPU. Other EGL based apps (eglgears for example) work as expected, also if desmume uses SDL for init then everything also works right.

PR #865, opened 2024-11-12
EGL: use eglGetPlatformDisplay if available

thesourcehim:
To properly use prime render offload eglGetPlatformDisplay must be used instead of eglGetDisplay. It's available in EGL 1.5, so use it if available, fallback to eglGetDisplay if not. Helps #864.


rogerman:
We can assume that the EGL v1.5 headers will be available at compile time, so no need for #ifdef here. (However, we are still requiring EGL v1.4 at runtime.)

Also, your name of "attr" isn't descriptive enough, since there are other attributes in that particular function. Please rename "attr" to "displayAttr", thanks!

thesourcehim:
I guess I just try all 3 calls to obtain display handle, I don't see a way to obtain runtime EGL version before calling eglInitialize.

PR #874, opened 2024-12-06
add support for mingw build

rofl0r:
this allows to easily cross-compile the windows gui app for 32 or 64 bits even.


zeromus:
as long as it doesn't break msvc. let me know when youre done

rofl0r:
should be complete from my side. you might want to try the aviout.cpp fix though; i'm not 100% sure it's correct, and avi recording crashes in the mingw build - though that might have other reasons. once it's merged i can set up a CI job

crashGG:
Hi,could you add mingw-based builds to the workflow?

rofl0r:
@zeromus so - is it ok for you to merge this? since there's currently some refactor stuff going on i fear that merging this will soon become a pain, so i guess it's best to do it ASAP while everything still works.

zeromus:
As long as it doesn't break msvc.. sure..

PR #891, opened 2025-03-11
Better approach for the Xbrz filters

FJRODDUR:
I've changed some parts of the Xbrz filters for better performance on low-end PCs. This really gave me a FPS boost on my benchmark PC :D (i5-3340 3.10GHz with integrated graphics). Feel free to ask questions or leave a comment :3

### DIFF
--- a/desmume/src/filter/xbrz.cpp
+++ b/desmume/src/filter/xbrz.cpp
@@ -478,7 +478,7 @@ const bool sPixDoLineBlend(const Kernel_3x3& ker, const char blend, const xbrz::
 		return true;
 	
 	//make sure there is no second blending in an adjacent rotation for this pixel: handles insular pixels, mario eyes
-	if (getTopR(blend) != BLEND_NONE && !sPixEQ<ColorDistance>(e, g, cfg)) //but support double-blending for 90¡ corners
+	if (getTopR(blend) != BLEND_NONE && !sPixEQ<ColorDistance>(e, g, cfg)) //but support double-blending for 90ï¿½ corners
 		return false;
 	if (getBottomL(blend) != BLEND_NONE && !sPixEQ<ColorDistance>(e, c, cfg))
 		return false;
@@ -550,7 +550,7 @@ void blendPixel(const Kernel_3x3& ker,
                 return true;
 
             //make sure there is no second blending in an adjacent rotation for this pixel: handles insular pixels, mario eyes
-            if (getTopR(blend) != BLEND_NONE && !eq(e, g)) //but support double-blending for 90° corners
+            if (getTopR(blend) != BLEND_NONE && !eq(e, g)) //but support double-blending for 90ï¿½ corners
                 return false;
             if (getBottomL(blend) != BLEND_NONE && !eq(e, c))
                 return false;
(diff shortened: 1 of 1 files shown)


desmume/src/filter/xbrz.cpp
```diff
@@ -478,7 +478,7 @@ const bool sPixDoLineBlend(const Kernel_3x3& ker, const char blend, const xbrz::
 		return true;
 	
 	//make sure there is no second blending in an adjacent rotation for this pixel: handles insular pixels, mario eyes
-	if (getTopR(blend) != BLEND_NONE && !sPixEQ<ColorDistance>(e, g, cfg)) //but support double-blending for 90� corners
+	if (getTopR(blend) != BLEND_NONE && !sPixEQ<ColorDistance>(e, g, cfg)) //but support double-blending for 90� corners
```
rofl0r, line 481: it seems your editor automatically replaces characters with your local windows charset resulting in bogus output for regular utf8 users. configure it so it doesn't touch stuff *you* didn't touch. but if you did it yourself: we don't want gratuitous whitespace or other changes in stuff unrelated to the PR's core functionality. that just produces noise for reviewers and later people reading the commit log.

desmume/src/filter/xbrz.cpp
```diff
@@ -550,7 +550,7 @@ void blendPixel(const Kernel_3x3& ker,
                 return true;
 
             //make sure there is no second blending in an adjacent rotation for this pixel: handles insular pixels, mario eyes
-            if (getTopR(blend) != BLEND_NONE && !eq(e, g)) //but support double-blending for 90� corners
+            if (getTopR(blend) != BLEND_NONE && !eq(e, g)) //but support double-blending for 90� corners
```
rofl0r, line 553: likewise.

desmume/src/filter/xbrz.cpp
```diff
@@ -781,260 +781,294 @@ void scaleImage(const uint32_t* src, uint32_t* trg, int srcWidth, int srcHeight,
...
+        uint32_t& pixel5 = out.template ref<2, 2>();
+        
+        alphaGrad<1, 4>(pixel1, col);
+        alphaGrad<1, 4>(pixel2, col);
+        alphaGrad<3, 4>(pixel3, col);
+        alphaGrad<3, 4>(pixel4, col);
+        pixel5 = col;
     }
-
+
```
rofl0r, line 891: likewise.

rofl0r:
what are the before/after numbers you measured? the commit message has ample space to document that, also to document what you changed to achieve the better performance.

FJRODDUR:
> what are the before/after numbers you measured? the commit message has ample space to document that, also to document what you changed to achieve the better performance.

Here is the benchmark, only two screenshots before and after the approach
**Before**
imagen
**After**
imagen

It is a 5-10% speed improvenment on my local machine

And I'm very sorry about the useless changes like the whitespaces, I'll change them as soon as possible, as well as the buggy charset.

zeromus:
The only optimization I see is changing this
A(x())
B(y())
to 
temp a = x()
temp b = y()
A(a)
B(b)

There is no possible way this yields a speedup in the general case, and I doubt it in the specific case.

It would be better to seek out someone else's more-recently-optimized filters

FJRODDUR:
> The only optimization I see is changing this A(x()) B(y()) to temp a = x() temp b = y() A(a) B(b)
> 
> There is no possible way this yields a speedup in the general case, and I doubt it in the specific case.
> 
> It would be better to seek out someone else's more-recently-optimized filters

Okay okay

PR #896, opened 2025-04-01
Update wifi.cpp - added faster reflect functions

gnudles:
I know this is so insignificant because it only hapens on initialization. But maybe future implementation might require it. Feel free to reject it.

### DIFF
--- a/desmume/src/wifi.cpp
+++ b/desmume/src/wifi.cpp
@@ -346,6 +346,24 @@ static u32 reflect(u32 ref, char ch)
 	return value;
 }
 
+static u32 reflect8(u32 ref)
+{
+	ref = ((ref & 0x0F) << 4) | ((ref & 0xF0) >> 4);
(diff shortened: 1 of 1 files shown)


rogerman:
Your replacement functions, reflect8() and reflect32() rearrange the bits on a per-nibble basis, but the original reflect() function rearranges the bits on a per-bit basis. Therefore, these are NOT equivalent functions at all. Nice try trying to optimize, but we've gotta reject this PR because your new functions cannot replace the old function. Optimization is hard.

rogerman:
Okay, might of analyzed this PR too quickly -- the functions actually are equivalent. My bad.

Still, at the end of the day, we're just filling a LUT of 256 32-bit values only once at initialization, so this isn't even a performance issue. Meh. Because the functions are, in fact, equivalent, I'll reopen this PR if anyone actually cares.

gnudles:
Yes I know this is not significant at all. I just thought, maybe one day, someone would need to use it in a different place in the code, so he better have a faster implementation...

Issue #897, opened 2025-04-07
Using touchpad (ds4/ds5) for direct touchscreen use.


backgamon:
Hi, 
Any chance for using the touchpad from DS4/DS5 on the NDS touchscreen like there(3DS emu):
https://github.com/azahar-emu/azahar/issues/746#issuecomment-2742739462

https://github.com/azahar-emu/azahar/pull/777
Maybe it can be converted for DesMume.

thesourcehim:
ds4/ds5 touchpad is interpreted as normal touchpad in the system. This means you can keep it pressed and move your finger to get the desired effect. Keeping it always pressed can be hard though.
I think I can process SDL touchpad events for detecting touch/lift in GTK frontends to make toughpad usage easier. Right now I'm quite busy, I'll take a look in a week or so.

thesourcehim:
@backgamon, in case you are using linux, please test PR #903

backgamon:
Unfortunately i’m under Windows 10 x64

zeromus:
Thanks for sharing the reference. Not easy because we don't use SDL, but it's inspirational and useful nonetheless

PR #903, opened 2025-04-19
Linux/GTK: add support for gamepad touchpad -> touch screen input

thesourcehim:
For this I had to somewhat rework SDL input handling and add game controller api usage: if joystick is recognized as a game controller by SDL I open it as a game controller, otherwise fallback to generic joystick open. This is required to receive touchpad events. Additionally I have to filter out joystick events if joystick is open as game controller, otherwise there will be double events.

Touchpad finger coords are translated into touchscreen coords thus making touchpad act like touchscreen. This may be what is wanted in #897, but I implemented it only for Linux GTK frontends. Cli frontend must process changes in touchpad_status structure for touchpad translation to work (just like it does for mouse_status structure).

Note that hidraw access permissions are required to use touchpad, i.e. /dev/hidraw* file corresponding to gamepad must be accessible for user.

### DIFF
--- a/desmume/src/frontend/posix/shared/ctrlssdl.cpp
+++ b/desmume/src/frontend/posix/shared/ctrlssdl.cpp
@@ -144,7 +144,9 @@ BOOL init_joy( void) {
            printf("Touchpads: %d\n", SDL_GameControllerGetNumTouchpads(gcont));
            printf("%s\n", SDL_GameControllerHasSensor(gcont, SDL_SENSOR_ACCEL) ? "Accelerometer present" : "Accelerometer not present");
            printf("%s\n", SDL_GameControllerHasSensor(gcont, SDL_SENSOR_GYRO) ? "Gyroscope present" : "Gyroscope not present");
-           open_joysticks[i]={TRUE, {.gcont=gcont}, SDL_JoystickInstanceID(SDL_GameControllerGetJoystick(gcont))};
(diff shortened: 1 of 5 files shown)


desmume/src/frontend/posix/shared/ctrlssdl.cpp
```diff
@@ -401,53 +589,49 @@ do_process_joystick_events( u16 *keypad, SDL_Event *event) {
...
       if (key != 0)
         ADD_KEY( *keypad, key );
       break;
-
-      /* Joystick button released */
-    case SDL_JOYBUTTONUP:
-      key_code = ((get_joystick_number_by_id(event->jbutton.which) & 15) << 12) | JOY_BUTTON << 8 | (event->jbutton.button & 255);
+    
+    /* Joystick button released */
+    case 3:
```
rofl0r, line 628: can you make this an enum ? then we dont need comments to explain magic values

thesourcehim: Done

rofl0r:
apparently the guy cant test it. do you have a touchpad device that you tested it with?

thesourcehim:
I tested with ps5 dualsense touchpad. Also I'm currently trying to build gtk3 frontend for windows, all dependencies seem to be available via vcpkg, so maybe I'll be able to provide executable for him to test.

PR #905, opened 2025-05-10
Remove OSMesa support

linkmauve:
Mesa has now removed OSMesa from its codebase, see https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/33836

There is no reason to keep this any longer, especially since EGL can do the same and more.


rofl0r:
why remove something that works ? just because it's no longer officially maintained doesn't mean people immediately stop using it.

rogerman:
Actually... I can attest to OSMesa never working. At least for me, I couldn't get OSMesa to work years ago, I couldn't get it to work in August 2024 when I was reworking all of the OpenGL context code, and I can't get it to work right now. When I reworked OSMesa last year, I marked it as deprecated for the following reasons:

1. AFAIK, OSMesa context creation has been in an untested state for several years since I couldn't get it to work myself, and I haven't found any info about other people getting it to work either.
2. In last year's OpenGL context rework, I brought back GLX specifically to fill the role of the old-school compatibility option because I couldn't get OSMesa to work, so there is no need for OSMesa to exist as a second old-school compatibility option.
3. Most users are expected to be creating their OpenGL contexts through SDL or EGL. I'm not aware of anyone who actually uses OSMesa in practice.

So in my opinion, I would support the removal of OSMesa. We can always bring it back if some OSMesa user actually wants it back, but I am highly doubtful of this. Otherwise, that user, even one user, would have already mentioned something about it after all these years.

rofl0r:
well, if it doesnt work at all, that's at least a valid reason to remove it. but "freedesktop said it's no longer a thing" isn't.

rogerman:
@rofl0r, correct. The determining factor in this case should be total unavailability rather than the state of support. In other words, only if a library becomes completely unavailable for download/install should it be removed from DeSmuME's code base. But I can see that libOSMesa.so is, in fact, available on apt as well as Synaptic Package Manager, so users can easily use OSMesa if they want to. Removing OSMesa support just because freedesktop dun-like-it-no-more is not a valid reason for its removal.

But still, I've never gotten OSMesa to work after all these years, and I haven't seen any others users say that they use it, and so it would be nice to get rid of OSMesa since GLX can easily take its place.

PR #942, opened 2026-01-21
puppy shop EU fix

rofl0r:


rogerman (commented):
You actually don't need to explicitly enable gamehacks.en by default, because the core code already does this:
https://github.com/TASEmulators/desmume/blob/f96a6d6e16a18a97f4eb696f8088b52a54d3a61f/desmume/src/NDSSystem.cpp#L3386

rofl0r:
rogerman: good find. i removed the offending commit.

rogerman:
Tested all three games, and they seem to work just fine with this game hack.

Here's a screenshot of the European version of My Puppy Shop (NTR-YPTP-EUR):
puppy_shop_works

Issue #483, opened 2021-11-02
Game freezes after logo


BlueTigress07:
Every time I try to load Rabbids Go Home it shows the ubisoft logo and freezes on a black screen with no audio. After it sits for a bit, In the second window it says “ARM9: undefined instructions: 0xEA000075 PC=0x02004000” . This is the only game that does this

 I looked around for answers but couldn't find any fixes. Im not great with these sorts of errors so a slightly more detailed suggestion or fix would be greatly appreciated

zeromus:
Works fine for me. Stop using a million years old version of desmume or cheating or patching, or else make sure your rom isn't corrupted.

zeromus:
That's useful information, but then why do my notes from 3 years ago say it worked just fine for me?

Sukotto-1999:
Welp, the game crashes for me, maybe should add a commit to emulate this load precision or implement a cheat code to get it working

Sukotto-1999:
Okay, adjusting is complex. Just for now, you either use DS-Scene Rom Tool with my updated custom database or use action replay code for the game to work.

US:
5203CF18 E28DD02C
0203CF18 EB0F0878
D2000000 00000000
023FF100 E59D1000
E23FF104 00000030
E3510001 1A000003
E28F1010 E891001E
E5813000 E5824000
E28DD02C E12FFF1E
0218D5C4 0218D5CC
E3A05921 E3A04000
D2000000 00000000

EUR:
5203D108 E28DD02C
0203D108 EB0EFCBC
D2000000 00000000
023FF100 E59D1000
E23FF104 00000030
E3510001 1A000003
E28F1010 E891001E
E5813000 E5824000
E28DD02C E12FFF1E
0218DA4C 0218DA54
E3A05921 E3A04000
D2000000 00000000

rofl0r:
> The game runs 100 rom reads of 0x4000 bytes and times how long will it take. The loading time must take around 80000 and 88000 ticks (152 to 168ms) or the game will hang before the title screen.

do you know the actual time it takes with desmume ?

taxicat1:
> do you know the actual time it takes with desmume ?

It consistently takes 0x7E046 ticks on US and 0x7E050 on EU. Only *slightly* too fast.

[...]

PR #943, opened 2026-01-22
use @taxicat1's patch for rabbids go home

rofl0r:
a way less intrusive patch for the timing bug experienced with rabbids go home.

ref #483


rofl0r:
works fine in my tests. note that testing this requires removal of existing .dsv battery saves.

PR #947, opened 2026-02-01
Windows: Add configurable fullscreen hotkey F11

RedBlackAka:
Add a configurable toggle fullscreen hotkey for Windows under Tools, that is set to F11 by default, following common practice among other applications (especially emulators). The Linux GTK build seemingly already uses F11 anyways. Does not remove the ALT + Enter hotkey, for those who would like to continue using the old one.

### DIFF
--- a/desmume/src/commandline.cpp
+++ b/desmume/src/commandline.cpp
@@ -67,7 +67,7 @@ static const char* help_string = \
 "                            multipler; 1:256x192 (default), 2:512x384," ENDL
 "                            3:768x576, 4:1024x768, 5:1280x960" ENDL
 " --windowed-fullscreen" ENDL
-"                            Launches in windowed fullscreen (same as alt+enter)" ENDL
+"                            Launches in windowed fullscreen (same as F11)" ENDL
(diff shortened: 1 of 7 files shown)


rofl0r:
alt-enter is the standard fullscreen shortcut for a reason. linux/X11 window managers already use F11 to maximize windows, so depending on the window manager the keypress is either swallowed by the WM and never reaches SDL, or processed by each side with incompatible behaviour.

RedBlackAka:
Ahh alright. Do you want me to remove the change for SDL?

rofl0r:
> Do you want me to remove the change for SDL?

yeah. i don't  use the windows port myself, but i think its users wouldn't like their fullscreen hotkey no longer working either. so why not just stick with alt-enter anywhere?

RedBlackAka:
Will be done. Though Alt + Enter is not removed on Windows. This simply adds F11 in addition, which is the standard for fullscreen on Windows, making usage more intuitive for unfamiliar users (and helping my muscle memory lol)

PR #948, opened 2026-02-04
features_cpu.c: only check ARM features on ARM

pkubaj:
This currently breaks build for me on my ppc64le workstation.


rofl0r:
could you check whether upstream libretro has fixed the same problem differently? your patch is most likely wrong for other platform/arch combos like e.g. freebsd or even windows ARM, since the check_arm_cpu_feature function only gets compiled in `#if defined(__linux__) && !defined(CPU_X86)`, so a correct patch would need to have the same conditional in both spots.

PR #958, opened 2026-04-04
fix: security, memory safety, correctness, and CI improvements

Alugor:
## Summary

- **Security hardening**: Fix buffer overflows in wifi (DNS, RX packets), GDB stub, CLI args, firmware LZ routines, and ZIP path handling. Replace unsafe `strcpy`/`sprintf`/`sockaddr` casts with bounds-checked alternatives.
- **Memory safety**: Fix memory leak in SPU_Init, GL shader cleanup on failure, GDB stub resource leaks, missing virtual destructor on `GameHackCheat`, and double `NDS_Reset()` in backup import.
- **Performance**: Guard MMU breakpoint scans with `empty()` check on all 6 read/write paths; restore `FORCEINLINE` on SPU hot path.
- **Correctness**: Implement `export_dsv()` for native `.dsv` save export (was silently failing), fix `INVALID_SOCKET` comparison on Windows (WiFi was broken), fix double reset after save import across GTK2/Windows frontends, add `SO_EXCLUSIVEADDRUSE` for GDB port reuse on Windows.
- **CI**: Add native macOS ARM64 build (Apple Silicon NEON-A64 SIMD coverage), fix deprecated/removed MinGW compiler flags (`-fforce-addr`, `-fstrength-reduce`, `-march=pentium4`), fix artifact name typo.
- **Platform**: Define `HOST_WINDOWS` for MinGW builds (`__MINGW32__`/`__MINGW64__`).

## Changed files

18 files changed, 287 insertions(+), 75 deletions(-) across `src/`, `.github/workflows/`, and frontends.

## Test plan

- [ ] Verify CI passes on all platforms (Linux, macOS Intel, macOS ARM64, MinGW)
- [ ] Test GDB stub attach/detach cycle with port reuse
- [ ] Test save import/export for `.dsv`, `.sav`, `.duc`, `.ssp` formats
- [ ] Test WiFi emulation on Windows (INVALID_SOCKET fix)
- [ ] Verify no regressions in normal emulation gameplay

### DIFF
--- a/desmume/src/commandline.cpp
+++ b/desmume/src/commandline.cpp
@@ -457,10 +457,10 @@ bool CommandLine::parse(int argc,char **argv)
 		CommonSettings.autodetectBackupMethod = autodetect_method;
 
 	//TODO NOT MAX PRIORITY! change ARM9BIOS etc to be a std::string
-	if(_bios_arm9) { CommonSettings.UseExtBIOS = true; strcpy(CommonSettings.ARM9BIOS,_bios_arm9); }
-	if(_bios_arm7) { CommonSettings.UseExtBIOS = true; strcpy(CommonSettings.ARM7BIOS,_bios_arm7); }
-	#ifndef HOST_WINDOWS 
-		if(_fw_path) { CommonSettings.UseExtFirmware = true; CommonSettings.UseExtFirmwareSettings = true; strcpy(CommonSettings.ExtFirmwarePath,_fw_path); }
+	if(_bios_arm9) { CommonSettings.UseExtBIOS = true; strncpy(CommonSettings.ARM9BIOS,_bios_arm9,sizeof(CommonSettings.ARM9BIOS)-1); CommonSettings.ARM9BIOS[sizeof(CommonSettings.ARM9BIOS)-1]='\0'; }
+	if(_bios_arm7) { CommonSettings.UseExtBIOS = true; strncpy(CommonSettings.ARM7BIOS,_bios_arm7,sizeof(CommonSettings.ARM7BIOS)-1); CommonSettings.ARM7BIOS[sizeof(CommonSettings.ARM7BIOS)-1]='\0'; }
+	#ifndef HOST_WINDOWS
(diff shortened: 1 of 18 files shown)


desmume/src/commandline.cpp
```diff
@@ -457,10 +457,10 @@ bool CommandLine::parse(int argc,char **argv)
 		CommonSettings.autodetectBackupMethod = autodetect_method;
 
 	//TODO NOT MAX PRIORITY! change ARM9BIOS etc to be a std::string
-	if(_bios_arm9) { CommonSettings.UseExtBIOS = true; strcpy(CommonSettings.ARM9BIOS,_bios_arm9); }
-	if(_bios_arm7) { CommonSettings.UseExtBIOS = true; strcpy(CommonSettings.ARM7BIOS,_bios_arm7); }
-	#ifndef HOST_WINDOWS 
-		if(_fw_path) { CommonSettings.UseExtFirmware = true; CommonSettings.UseExtFirmwareSettings = true; strcpy(CommonSettings.ExtFirmwarePath,_fw_path); }
+	if(_bios_arm9) { CommonSettings.UseExtBIOS = true; strncpy(CommonSettings.ARM9BIOS,_bios_arm9,sizeof(CommonSettings.ARM9BIOS)-1); CommonSettings.ARM9BIOS[sizeof(CommonSettings.ARM9BIOS)-1]='\0'; }
```
rofl0r, line 460: strncpy isn't safer, but slower:

```
     The  stpncpy()  and  strncpy() functions shall copy not more than n bytes
     (bytes that follow a NUL character are not copied) from the array pointed
     to by s2 to the array pointed to by s1.

     If  the  array pointed to by s2 is a string that is shorter than n bytes,
     NUL characters shall be appended to the copy in the array pointed  to  by
     s1, until n bytes in all are written.
```
from man 3p strncpy.
so it always writes the full bufsize, and if the string to copy is longer then it won't get zero-terminated, allowing consumers to read out of bounds, causing UB.
the safe alternative is `snprintf(copybuf, BUFSIZE, "%s", orgstring)` since it will at least zero-terminate the copy if it's longer than BUZSIZE.

rofl0r:
basically not opposed to this, but each change should be a separate PR so it can be properly discussed and reviewed for its own merits. you also seem to change unrelated stuff like changing the target settings for the mingw build. reasoning behind it? the build was specifically added so people using older hardware and OS combos can still enjoy an up-to-date desmume build that works.

Alugor:
Splitting into separate PRs per review feedback. Thanks for the input!

PR #963, opened 2026-04-04
fix: backup save export/import correctness

Alugor:
## Summary

**Backup export:**
- Implement `export_dsv()` for native DeSmuME `.dsv` save export (`exportData` previously silently failed for `.dsv` despite API documentation)
- Add `padSize` validation and `malloc` NULL check in `import_dsv` (prevents crash on malformed `.dsv` files with huge/zero padSize)
- Replace `strcpy` with `strncpy` in `exportData` `.sav` path handling
- Fix documentation to list all supported export formats

**Double NDS_Reset fix (all frontends):**
- Remove redundant `NDS_Reset` after `importData` in GTK2 and Windows frontends (`importData` already calls `NDS_Reset` internally on success)

## Files changed

- `desmume/src/mc.cpp`, `desmume/src/mc.h`
- `desmume/src/frontend/interface/interface.cpp`, `interface.h`
- `desmume/src/frontend/posix/gtk2/main.cpp`
- `desmume/src/frontend/windows/importSave.cpp`

## Test plan

- [ ] Test save import/export for `.dsv`, `.sav`, `.duc`, `.ssp` formats
- [ ] Test importing a malformed `.dsv` with bad padSize
- [ ] Verify single reset after save import on GTK2 and Windows

### DIFF
--- a/desmume/src/frontend/interface/interface.cpp
+++ b/desmume/src/frontend/interface/interface.cpp
@@ -293,9 +293,8 @@ EXPORTED BOOL desmume_backup_import_file(const char *filename, unsigned int forc
 	if (!success) {
         return FALSE;
 	}
-    
-    NDS_Reset();
-	
+
+	// NDS_Reset() is already called inside importData() on success
     return TRUE;
 }
 
--- a/desmume/src/frontend/interface/interface.h
+++ b/desmume/src/frontend/interface/interface.h
@@ -131,7 +131,12 @@ EXPORTED char* desmume_savestate_slot_date(int index);
 EXPORTED BOOL desmume_backup_import_file(const char *filename, unsigned int force_size);
 
 /**
- * Export current battery save to .dsv file.
+ * Export current battery save to file.
+ *
+ * Supported formats (determined by file extension):
+ *   - .dsv  - DeSmuME native save format (raw data + metadata footer)
+ *   - .sav  - Raw save data (no metadata)
+ *   - .sav* - no$GBA format
  *
  * Prerequisites:
  *   - MMU_new.backupDevice is a global object initialized at program startup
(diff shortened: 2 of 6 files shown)


rofl0r:
> Replace strcpy with strncpy in exportData .sav path handling

did you see my comment about strncpy in your original PR ?

rofl0r:
https://github.com/TASEmulators/desmume/pull/958#discussion_r3035550628

really seeing strncpy used anywhere is a real code smell. it means people blindly copy/paste from bogus tutorials without reading the function's manpage.

zeromus:
This is someone using an AI agent. They may not even know what you're talking about. I don't want hit and run commits from them. Anyone willing to get ramped up and stay invested can run the same tools, obtain the same findings, and more importantly stick around to work through the fallout.

Alugor:
I'm fixing actual security vulnerabilities here — buffer overflows in wifi, heap overflows in the GDB stub, resource leaks, broken WiFi on Windows due to unsigned socket comparisons. These are real bugs that affect real users.

Instead of reviewing the code on its merits, you're speculating about how it was produced and preemptively dismissing it. That's not a code review, that's a baseless accusation.

I'm right here — I split the original PR into six separate ones because rofl0r asked for it, and I'm about to push fixes for the strncpy issue he correctly pointed out. That's not "hit and run," that's exactly how open source collaboration is supposed to work.

If you see a technical problem with any of these changes, I'm all ears. But if the only objection is a hunch about my workflow, I'd ask you to reconsider whether that's a productive use of everyone's time.

rofl0r:
even though you split the PRs by topic, it's still several changes in each one. as soon as you need a bullet point list to list changes that's a clear indicator that a PR does too much. for example removing gratuitous NDS_reset calls deserves its own PR, just like addition of a new dsv export func does.

Jules-A:
> even though you split the PRs by topic, it's still several changes in each one. as soon as you need a bullet point list to list changes that's a clear indicator that a PR does too much. for example removing gratuitous NDS_reset calls deserves its own PR, just like addition of a new dsv export func does.

You are wasting your time arguing with an AI agent, not a human. That response is the exact same one AI agents give when their PRs get denied.

>Anyone willing to get ramped up and stay invested can run the same tools, obtain the same findings, and more importantly stick around to work through the fallout.

Not only that but the agents usually produce worse output than running them through the full model.

PR #967, opened 2026-05-10
Cocoa Port: Input mapping improvements (toggle speed, toolbar speed display, status text)

SpyrosBou:
## Summary

Three focused improvements to the macOS Cocoa port's input/speed UX.

### 1. Add `Set Speed (Toggle)` input binding command
A toggle counterpart to the existing hold-style `Set Speed` command. Multiple bindings supported, each with its own scalar via the existing per-binding scalar slider sheet (no XIB changes).
- Pressing the same toggle binding again disengages and restores the previous speed.
- Pressing a *different* toggle binding while engaged switches scalar in one press (no intermediate restore).

### 2. Toolbar Speed item shows current speed; add 4x/6x/8x menu entries
- Toolbar `Speed` item now reflects the current execution speed (rounded to nearest int, clamped 1–8x) when the limiter is enabled. Clicking advances 1x → 2x → 4x → 6x → 8x → 1x.
- When the speed limiter is disabled, the toolbar item shows `Unlimited`.
- Adds 4x / 6x / 8x entries to `Emulation > Execution Speed` programmatically at startup. The `Custom` item state recognises the new tagged presets.

### 3. Show bound NDS button name in input status text
The bottom status bar now reads e.g. `Keyboard: 4 - B` with the bound command name (`B`) in **bold** when an input has a mapping, making it easier to see which console button or action a key is mapped to.

To support attributed strings while preserving all existing plain-string status messages, `statusText` was loosened from `NSString *` to `id`. The `value:` binding in `DisplayWindow.xib` continues to render both forms correctly.

## Test plan
- [x] Builds clean: `xcodebuild -scheme "DeSmuME (macOS App; AppleSilicon -- Latest Xcode)" -configuration Release` (Xcode 16, macOS 15 / Apple Silicon).
- [ ] Bind keys to `Set Speed (Toggle)` at multiple scalars (2x and 4x). Verify single-press transitions: 1x → 2x; 2x → 4x (one press); 4x → 1x (same-binding press disengages); 4x → 2x (different binding while engaged).
- [ ] Toolbar `Speed` item: cycle 1 → 2 → 4 → 6 → 8 → 1 by clicking. Verify correct integer label at each.
- [ ] `Emulation > Disable Speed Limit`: toolbar shows `Unlimited`.
- [ ] `Emulation > Execution Speed` submenu shows new 4x / 6x / 8x entries; checkmark follows the active preset.
- [ ] Press a key bound to a console button (e.g., `Z` → `B`): bottom status bar reads `Keyboard: Z - B` with `B` in bold.
- [ ] Press an unbound key: status reads `Keyboard: <key>` with no trailing tag.
- [ ] Touchscreen click in `Main`-only mode: `Mouse: Left - Touch`. In dual-display mode: `Mouse: Left X:.. Y:.. - Touch`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rogerman:
@SpyrosBou, thanks for the PR! I tested your changes extensively, and have come to like the general direction in which you have tried to make improvements to DeSmuME. Unfortunately, I will have to reject this PR due to it bringing about certain inconsistencies to the user experience. However, I have reimplemented all of the features you have brought forth using a better code implementation as well as maintaining a consistent user experience, using this PR as inspiration. See commit a757047 for more details.

Below is the rationale I have used in the rejection of this PR.

`1. Add Set Speed (Toggle) input binding command`
`A toggle counterpart to the existing hold-style Set Speed command. Multiple bindings supported, each with its own scalar via the existing per-binding scalar slider sheet (no XIB changes).`
This is a nice feature to have, but the better implementation is not to have a completely separate, but similar, command to Set Speed, but instead to have the same Set Speed command be configurable for either 'hold-to-toggle' or 'toggle-and-latch' control. This is similar to how the various NDS controls (Up, Down, Left, Right, A, B, X, Y, etc.) are configurable for turbo, rather than having each NDS control have separate turbo and non-turbo commands.

`• Pressing the same toggle binding again disengages and restores the previous speed.`
That part about 'restores the previous speed' is a very important feature. The 'previous speed' is set in the Execution Control panel, when the user changes the slider (or clicks the Normal Speed button). All the speed toggling commands depend on this.

Speed toggling is considered to be a temporary change, never a permanent one. This idea applies to whether you 'hold-to-toggle' or 'toggle-and-latch'. For example, if I configure my gamepad to use two of its shoulder buttons for setting the speed to 1.0x and 2.0x, respectively, and then I manually set the speed in Execution Control to 0.5x, then I definitely want each shoulder button to restore my speed to 0.5x every time I toggle. I don't want to lose my previously set speed while I'm trying to do some careful in-game maneuvers with the other buttons -- I want my speed toggling to work with the previous speed that I set.

`Pressing a different toggle binding while engaged switches scalar in one press (no intermediate restore).`
A nice quality-of-life improvement. If previous versions of DeSmuME didn't already do this, then that would be considered a bug.

`2. Toolbar Speed item shows current speed; add 4x/6x/8x menu entries`
`• Toolbar Speed item now reflects the current execution speed (rounded to nearest int, clamped 1–8x) when the limiter is enabled. Clicking advances 1x → 2x → 4x → 6x → 8x → 1x.`
`• When the speed limiter is disabled, the toolbar item shows Unlimited.`
This behavior is inconsistent with how all the other toolbar buttons work, as well as being a basic UI error (and violating Apple's UI guidelines). Toolbar buttons should NEVER reflect the current state -- they should ALWAYS show the action that the user will take to change the state to. In other words, if a toolbar button says "Unlimited", then I as the user would expect to click it to disable the speed limiter. Your implementation doesn't do that, which is confusing. Even worse, clicking the button while the speed limiter is disabled won't show any user feedback in the toolbar, making the user think the button might be broken. This furthers the confusion.

It is possible for the user to present the toolbar with icons only. If the user selects this option, then there is no user feedback in the toolbar. New icons are needed here in order to show proper user feedback.

Furthermore, your implementation removes the current 2x → 1x toggle and replaces it with your 2x → 4x → 6x → 8x → 1x toggle. I'm sure many users would not appreciate this change. Instead, an additional toolbar item is needed so that both styles of speed toggles may be used.

`• Adds 4x / 6x / 8x entries to Emulation > Execution Speed programmatically at startup. The Custom item state recognises the new tagged presets.`
Hmmm... I don't see any Emulation > Execution Speed menu option. I made the test build directly from your fork, so I don't know...?

`3. Show bound NDS button name in input status text`
`The bottom status bar now reads e.g. Keyboard: 4 - B with the bound command name (B) in bold when an input has a mapping, making it easier to see which console button or action a key is mapped to.`
This is a great idea! But it shouldn't be limited to just keyboard presses and mouse clicks -- this feature should be universal for all inputs, including HID inputs.

`To support attributed strings while preserving all existing plain-string status messages, statusText was loosened from NSString * to id. The value: binding in DisplayWindow.xib continues to render both forms correctly.`
Don't like the unnecessary use of id, which is way too loosey-goosey a data type for my tastes. It's like using void* in C. If you want to use statusText as an NSAttributedString, then just change statusText to be an NSAttributedString. Doing this is fine, since I can already see some other use cases for statusText being an NSAttributedString beyond simply reporting some command mappings.

Commit a757047 should address all of these concerns. If commit a757047 doesn't provide all the features that this PR was asking for, then feel free to submit another issue (or better yet, another PR that describes what you're wanting), and I'll look into it.

SpyrosBou:
Thanks for the response, all notes very valid - This implementation was extremely rushed and not properly validated, I honestly just made the quick few edits for what functionality I wanted to see on the emulator myself - and opened the PR as a preview of potential ideas to curate before integrating (you ended up taking it that way so I'm glad you caught the intent) - On second thought, it should have been an issue/feature request and not a PR. 

I appreciate the detailed answer though, and was not expecting you would take the time to write all that down.

PR #975, opened 2026-07-29
savestate: bound firmware data when loading

acts-1631:
A crafted savestate can declare more firmware bytes than the fixed firmware
buffer holds. Loading it writes past the buffer and can corrupt emulator
memory.

Validate the firmware length against the buffer and remaining MMU chunk, and
fail incomplete reads.


zeromus:
already have a PR for this

PR #976, opened 2026-07-29
savestate: bound declared state lengths

acts-1631:
A crafted savestate header can request multi-GiB allocations before the loader
checks whether the file contains the declared payload. This can terminate the
emulator with a small input file.

Validate the file layout and cap compressed and uncompressed state sizes before
allocation.


zeromus:
already have a PR for this

PR #977, opened 2026-07-29
firmware: reject files larger than firmware storage

acts-1631:
The external firmware reader accepts 512 KiB V2 files although current
firmware storage is 256 KiB. The subsequent copy uses the 512 KiB size,
reading and writing beyond both objects.

Reject unsupported firmware files that exceed `NDSFirmwareData` storage before
the copy.


zeromus:
already have a PR for this

PR #978, opened 2026-07-29
gdbstub: bound memory-read reply size

acts-1631:
Bounds GDB memory-read replies before hex encoding them.

A peer connected to an enabled GDB stub could request a response larger than its fixed output buffer, corrupting emulator memory. The patch rejects oversized requests and packet payloads before framing them.


zeromus:
we don't need obscure bugfixes to rarely-used components from robots.

